PR #273, opened 2014-11-18
Experimental support for geteventstore.com as transport

floekke:
Very much work in progress!
Making a pull request as suggested to get some feedback on early stages :)


mookid8000:
not sure about the status for this - I'm closing it for now, as it will probably require quite a few changes to work as a transport for Rebus2

(which should be much easier though - Rebus2's `ITransport` interface is pretty simple)

floekke:
I dropped the implementation because of two things.

1. I got hooked on reactive extensions which offers a more functional way of composing events compared with rebus 'void Handle'. Which I think in comparison lacks in expressivity. 
2. At the moment, to my knowledge, rebus lacks a way of starting a subscription from some offset. E.g. from event or message number 1000.. This is a critical aspect of a log centric architecture..
Did I mention I think rebus is a really well polished library :)

mookid8000:
Ah, I see.... well, Rebus isn't really that well-suited to be an "event processor"... don't know if that's really a thing, but when you say "composing events" and "start a subscription from some offset" I'm thinking about correlating things in time, consumer-driven subscriptions, and event replaying, like you would do in systems based on event sourcing.

I've never used that kind of event processing to integrate systems - I've only used it within an application as a part of that application's event sourcing and CQRS architecture.

Did you take a look at [Cirqus](https://github.com/d60/Cirqus)? Cirqus has a command processor that processes commands, uses aggregate roots to do things and emit events, and then there's a complete subsystem of event processing in the form of asynchronous persistent views. The view subsystem has some nice features, e.g. arbitrary partitioning of the event stream, deterministic replay, automatic catch-up, etc., and both events and views can be stored in SQL Server, MongoDB, PostgreSQL + more.

floekke:
You're absolutely right about the relation to event sourced systems. But
the usefulness of consumer driven subscriptions extend beyond event sourced
systems. Instead of 'replaying old events' it might be 'attach a new
subsystem and process old messages', with no concept of aggregates being
involved..

I'm probably preaching to the choir here but if you haven't heard of kafka
and linkedins log centric architecture, heres some nice material:
https://engineering.linkedin.com/distributed-systems/log-what-every-software-engineer-should-know-about-real-time-datas-unifying

Anyway, once you have had a taste of a log centric architecture, its hard
to go back to a system where messages disappear :)

Regarding 'composing events' and 'event processor' I think sagas in rebus
are a subset of just that. Same goes for 'void Handle' .. 'Void handle'
reminds of the problems of composing event handlers that several reactive
extensions tutorials use as an example of the power of the framework..

I'll take another look at cirqus :)

mookid8000:
Now I understand what you're after :)

And yes, I knew about Kafka (at least the name and the concept), and I can definitely see how it makes it easier to introduce new event-driven applications.

Personally, I haven't really experienced any problems doing that, but maybe it's because I've often worked with systems where we'd introduce new integrations between them at a much much lower rate than we'd, say, introduce new views/change existing views in our CQRS architecture.

Question is whether I do it at such a slow rate because it's hard? I bet using something like Kafka will change the decisions you make because it removes many of the pains in those scenarios...?

mookid8000:
Btw - it's not really a replacement for something like Kafka, but Rebus has a built-in auditing feature whereby all successfully handled messages and all published messages are forwarded to a queue somewhere.

I'm planning on building an archiver that can take messages out of that queue and save them to a big database. One of the planned features is an event replaying function that can be used to replay published events and have them targeted at one specific (probably new) subscriber. I've take a few small steps in that direction, the most recent being [Station](https://github.com/rebus-org/Station), but it doesn't really do that much at the moment, and I think I need to take a few more stabs at the archiver thing before I get something that is right.

floekke:
In our case it looks like by introducing event streams as the main
integration mechanism, our services often don't need to expose any rpc like
api for queries.. In some cases consuming services might build a persistent
view themselves (with just the data they need!) or they might just react to
an event in the typical rebus/msmq fashion :)
It's an extra degree of freedom..
Because its so easy we will probably end up, dare i say it, with a
microservice architecture..
It's funny how kafka and event store are essentially the same thing, the
people behind just have different backgrounds. The kafka people use
terminology from distributed systems theory whereas the event store
terminology is more influenced by greg youngs whole ddd/cqrs/event sourcing
mind set.. I'm not sure thats a good thing for event store, they might be
limiting themselves :)

PR #281, opened 2015-01-02
Subscriptions - fixed throw-catch based subscription strategy

tobiaxor:
This is a normal situation (usually happens every time handler gets up), so shouldn't be handled by an exception.


mookid8000:
looks like it's only those silly Azure tests that are failing - I'll merge it right away :)

mookid8000:
It'll be out in 0.75.2 in a few seconds - thanks! :+1:

PR #282, opened 2015-01-10
Added return to source queue for Azure Service Bus transport.

seankearon:
This has added return to source for Azure Service Bus queues, including updating the RTSQ build.  I've lightly tested this against Azure, but haven't run the project's tests as my environment doesn't support them all.  This change set does not affect any existing code (excepting the build file).  Believe there's a build server with all set up, but not sure if I can access that.


seankearon:
Seems the build can't set up the test Azure SB queues.  Is that related to my pull request?

mookid8000:
it's not related to your PR :) thanks for taking the time to do this, I'll get it merged when I have the time... possibly later today

seankearon:
Cool, didn't think it was.  Hey - thanks for taking the time to make Rebus! :)

mookid8000:
excellent, thanks! :)

(⌐■_■)

Issue #248, opened 2014-07-22
Subscribe to all messages handlers from current assembly


pruiz:
Hi,

I cant seem to find an existing method to subscribe to all IHandleMessage implementations from an assembly (or from the current one). So I normally use something like:

	public static class SubscriptionExtensions
	{
		private static MethodInfo _busSubscribeMethod = typeof(IBus).GetMethod("Subscribe");

		private static IEnumerable<Type> GetMessagesHandledBy(Assembly assembly)
		{
			return assembly
					.ExportedTypes
					.SelectMany(x => x.GetInterfaces())
					.Where(i => i.IsGenericType && i.GetGenericTypeDefinition() == typeof(IHandleMessages<>))
					.Select(x => x.GetGenericArguments()[0])
					.ToArray();
		}

		public static void SubscribeFrom(this IBus bus, params Assembly[] assemblies)
		{
			foreach (var assembly in assemblies)
			{
				foreach (var msg in GetMessagesHandledBy(assembly))
				{
					var _subscribeMethod = _busSubscribeMethod.MakeGenericMethod(msg);
					_subscribeMethod.Invoke(bus, new object[] { });
				}
			}
		}

		public static void SubscribeFromCurrentAssembly(this IBus bus)
		{
			bus.SubscribeFrom(Assembly.GetCallingAssembly());
		}
	}

Would it make sense to add this methods to Rebus's core? I can send a pull-req if so.

Regards
Pablo

mookid8000:
Assembly scanning can be had now (since 0.76.0) - it's in the `Rebus.Extensions` package :) 

(allows you to go `bus. SubscribeByScanningForHandlers(anAssembly)`)

PR #283, opened 2015-01-11
Assembly auto subscribe

seankearon:
Added the ability to automatically scan assemblies for message handlers and to add corresponding subscriptions.  Looks like this:

    bus.SubscribeByScanningForHandlers(assembly1, assembly2);

### DIFF
--- a/src/Rebus.Extensions/Properties/AssemblyInfo.cs
+++ b/src/Rebus.Extensions/Properties/AssemblyInfo.cs
@@ -1,0 +1,36 @@
+﻿using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("Rebus.Extensions")]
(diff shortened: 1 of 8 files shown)


mookid8000:
Right - thanks for your contribution :) but I think this particular feature is better implemented in a separate project so as to avoid needlessly diluting the core APIs.

I've discussed this partciular feature before with @pruiz in #248 where I suggested he put the feature in a `Rebus.Conventions` or `Rebus.AssemblyScanning` project or something like that.

I'll be happy to accept your PR if you do that :) don't worry about writing the .nuspec, I'll make sure that the package gets published with all the other Rebus things.

What do you think?

seankearon:
Sure, makes sense! :)  New update just pushed.  I've created a Rebus.Extensions project and put the `bus.SubscribeByScanningForHandlers(...)` in there as an extension method.  Hope that naming is okay?  

I've left the test in TestRebusBus.cs.  Probably not quite the right place for it as the test exercises an extension, but starting a new fixture felt like littering! :)

Let me know what you think.

mookid8000:
your contribution is available in 0.76.0 which is on NuGet.org shortly - thanks! :+1:

seankearon:
Lovely job, thanks! :)

PR #288, opened 2015-01-23
Makes PoisonMessageInfo:.ctor public

### LINKED ISSUE #287: IErrorTracker can't be implemented with internal PoisonMessageInfo constructor
Hi,

First off, Rebus has been an absolute joy to work with. Everything is very well thought out. Thank you. :+1: .

I did come across one small issue: I'm trying to implement a custom `IErrorTracker`. However, I got down to the part where I've needed to implement `GetPoisonMessageInfo`:

        public PoisonMessageInfo GetPoisonMessageInfo(string id)
        {
            return null;
        }

Currently, there's no way to create an `PoisonMessageInfo` because the constructor is internal.

My main reason for creating my custom error tracker is to deal with a possible problem with distributed competing consumers running on different compute nodes. Each node will compete for messages by pulling messages from a NoSQL store. The current `ErrorTracker` uses a `ConcurrentDictionary`local to a running in a process. Since worker processes are running on different compute nodes, error tracking is not shared. So I have a needed to push the error tracking state into the (wrapped) transport message itself otherwise there is a potential to have a message bounce around different compute nodes (worst case) `N x 5` retries on an exception.

I can submit a pull-request to get `PoisonMessageInfo` a public constructor. Let me know.

Thanks,
Brian

bchavez:
* Makes PoisonMessageInfo:.ctor public
* Minor XML doc typo in IErrorTracker.cs.
Fixes #287

### DIFF
--- a/src/Rebus/Bus/IErrorTracker.cs
+++ b/src/Rebus/Bus/IErrorTracker.cs
@@ -46,7 +46,7 @@ namespace Rebus.Bus
         string ErrorQueueAddress { get; }
 
         /// <summary>
-        /// Retrieves the poison message information collected so far for the message with the specfied id.
+        /// Retrieves the poison message information collected so far for the message with the specified id.
         /// </summary>
         PoisonMessageInfo GetPoisonMessageInfo(string id);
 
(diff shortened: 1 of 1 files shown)


mookid8000:
your fix will be available in 0.77.1 on NuGet.org in a few minutes - thanks :+1:

bchavez:
Absolutely fantastic. Rebus is excellent. The source code and architecture is easy to follow and understand. Rebus is zero-friction and a joy to work with. Thank you for your help! :+1:

mookid8000:
thanks - your kind words make me blush :smiley: :blush:

PR #289, opened 2015-01-24
Added new IDetermineMessageOwnership implementation for using existing i...

DixonD-git:
...mplementations as chain-of-responsibility

### DIFF
--- a/src/Rebus.Extensions/DetermineMessageOwnershipFromOtherDeterminers.cs
+++ b/src/Rebus.Extensions/DetermineMessageOwnershipFromOtherDeterminers.cs
@@ -2,7 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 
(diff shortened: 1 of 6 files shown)


mookid8000:
Hi @DixonD-git , thanks for your contribution! I'm not going to add it to Rebus core though, because I want to avoid adding too many configuration options and ways of doing things to the core in order to keep it simple and focused.

I do appreciate your work though :), and I'll be happy to accept your PR if you move it to the [Rebus.Extensions](https://github.com/rebus-org/Rebus/tree/master/src/Rebus.Extensions) project which I think would be the appropriate place for this.

DixonD-git:
Hi @mookid8000 ! I have moved the code to Rebus.Extensions as you requested.

mookid8000:
excellent! it's available in 0.78.0 on NuGet.org now

PR #290, opened 2015-01-27
Idempotent Sagas extensions for Rebus

pruiz:
Hi,

Currently when an unexpected error is thrown (ie. connection to broker is lost) after a saga handler has been invoked and it's saga data has been persisted, but the message has not been acknowledged at/by the broker, the message will be reprocessed later on as expected. However, in such situations manual (and sometimes hard to maintain) defensive coding has to be applied to avoid duplicating side-effects from handling a message twice.

This pull-req add a pseudo-automatic Idempotent sagas feature which allows, by reusing (and extending) existing sagas mechanism, so in cases where a saga handler has been invoked, and it's data has been persisted, but the actual message was not acknowledged, rebus will avoid invoking (already invoked) sagas  and automatically re-play any previously generated side-effects (ie. bus transmissions/publishes).

The general idea behind this feature is being able to provide a way to automatically capture any message sent by user code invoking 'IBus' during a saga's handle method, so we serialize and save those messages as part of a saga's data bucket. This way, once a saga has been persisted, so are it's emitted messages (side effects).. And when time comes to re-invoke an already processed handler for an existing saga (with existing side-effects), we will be skipping such handler and simply re-send all messages to network.

In order to be backwards compatible and avoid breaking any existing code, this feature only applies to sagas inheriting from 'IdempotentSaga', which involves implementing/extending your SagaData with IIdempotentSagaData. This interface exposes a new class 'IdempotentSagaResults' which holds a serialized copy of the incoming message (just for diagnosing purposes) and a list of any emitted 'side-effects'.

Also, a future benefit from this works may be being able to avoid the restriction of having only one single Saga handler for a given message when a persister does not implement ICanUpdateMultipleSagaDatasAtomically (see note below). As when using this solution, when two or more IdempotentSagas are being executed, we can recover from an unexpected error happening in the middle of the execution pipeline.

This has been implemented as an independent project, but we needed to add a couple of hooks within Rebus's internals, so sending the whole feature (the IdempotentSagas project, our unit tests, etc.) in one single pull-req seemed the best way forward. However, as always, we are undoubtedly open to suggestions and improvements, in order to get this merged into mainline. ;)

Regards
Pablo

NOTE: I've not removed the current check against ICanUpdateMultipleSagaDatasAtomically, as I do not see right now how to do it only when all registered (ie. handlers to be invoked) are of IdempotentSaga type, w/o introducing a dependency on IdempotentSaga into Rebus' main assembly. :?


mookid8000:
wow,  that is an impressive piece of work! :)

Overall, I'm very positive about this solution - I've pondered a few time myself whether it would be possible to implement a truly idempotent saga, and it seems like you've accomplished this.

There's a few things that I'm less positive about, and that's the additions to Rebus core. I'm not sure there is any other way to do what you're doing though, so I guess that it is allright :) in particular, I'm not sure about the `SkipHandler`/`DoNotSkipHandler` thing on `IMessageContext`...... so I'm curious whether we could come up with a more elegant way of skipping one particular handler?

There's a slight overlap with the [IInspectHandlerPipeline](https://github.com/rebus-org/Rebus/blob/master/src/Rebus/IInspectHandlerPipeline.cs) service which would probably be able to provide the service you require.

I'm thinking that maybe you could install a decorator over `IInspectHandlerPipeline` which could be used to skip certain handler types, simply by filtering them out of the pipeline.

pruiz:
Hi,

Yeah, I wasnt too glad regarding SkipHandlers either, but initially it was a 'boolean' property (DoNotHandle) which was set at BeforeHandling, and reseted by Dispatcher/AftarHandling... And of course that was much worst.. ;)

I digged a bit at IInspectHandlerPipeline, but from what I saw, it is called early on by Dispatcher using something like (inspecthandlersPipeline.ToArray()) so it is computed with 'eager loading', and as such it wont defer evaluation until the point where we sure know that we want to skip a handler.. :?

Maybe we can modify it somehow to fit our purposes w/o breaking existing use cases, I just need a bit of enlightenment from your side.. ;)

mookid8000:
oh yeah, of course - there's a little bit of chicken-and-egg theme in this, because right now the order of operations is:

1. generate pipeline of handlers
1. iterate through handlers
1. if the handler is a saga handler, load the saga

which means that my proposed solution is not possible, because the saga instance (and thus the knowledge that this particular msg has already been processed) is not loaded until right before dispatch....


hmm... gotta think about this some more :)

pruiz:
Yeah, that's the issue I was refering too.. ;)

We thought about a few alternates (like dynamically generating proxy clases of SagaHandlers so we could 'proxy' the Handle() method and conditionally avoid running the real instance's Handle() when needed.. But that looked even more complex and not really worth the effort.

pruiz:
Hi Mogens,

Did you had a chance of thinking about this? ;)

Regards

mookid8000:
Yeah well.... still thinking! :) I have this feeling that there's a better way to indicate that a handler must be skipped, and I'm sure that way will become clear to me...

pruiz:
Hi,

No news on this front? ;)

Just let me know if there some kind of experiment you want me to try.. heh

mookid8000:
Hi @pruiz , sorry about the slow reaction on my part - I still haven't come up with a better solution than yours - but I haven't really had any time to look at the code really... I'll do it, I promise - thanks for your patience :)

pruiz:
ping.. ;)

mookid8000:
Thanks for the ping - I'll see if I can integrate your solution now :)

mookid8000:
excellent work @pruiz, and thank you for your patience :) I've finally gotten around to merge it and will release your contribution in 0.81.0 which is on NuGet.org in a few minutes :+1:

Issue #291, opened 2015-01-29
Async overloads of all IBus methods that send messages


mookid8000:
E.g. so you can

```
await bus.SendAsync(someMessage);
```
and
```
await bus.Defer(5.Minutes(), someMessage);
```
etc.

DixonD-git:
Hi @mookid8000! I started working on this in feature-send-async branch in my fork. However, I need advice. I've got a bit stuck on ITransactionContext.DoCommit when there are both synchronous and  asynchronous operations in the transaction scope now. Should it become ```event Func<Task> DoCommit;``` ? What's about AmbientTransactionContext.Commit, for instance, should it become async?

mookid8000:
Hi @DixonD-git - I don't understand what you're saying :) but I'd like to check out your branch when I have the time. Could you maybe open a PR and then we'll keep the discussion in there?

asgerhallas:
If you send a message async, does that mean that it will be sent before the transaction is committed? If not, is it then just for serialization and a bit of queue IO to be parallized?

(I moved this comment here, as I find it is more related to the feature as such)

mookid8000:
good point - right now afaik there's only two transport layers that support async: SQL Server and Azure Service Bus.

SQL Server could work because there's a transaction and every message is actually sent to the server when you call `bus.Send` - therefore, `bus.SendAsync` would end up in an `await sqlCommand.ExecuteNonQueryAsync()` or something like that.

Azure Service Bus, however, does not have transactions - therefore, to avoid messages "escaping" the current unit of work, sent messages are just put in a list of outgoing messages. And then, when the transaction commits, the messages are sent.

I realize now that the async send would be nicer if we can both `await bus.SendAsync()` (which the user would have to remember to do) and `await transactionContext.CommitAsync();` (which would just be the way Rebus would do it).

mookid8000:
closing this one for now - `async` / `await` will be first-class citizens in Rebus2 :)

asgerhallas:
I'm so excited :) when will be on Rebus2 nuget?

asgerhallas:
= when will Rebus2 be on nuget? :)

dalager:
Ah. That sounds nice!
await Rebus2 then.

PR #292, opened 2015-02-01
SendAsync support 

DixonD-git:
Related to https://github.com/rebus-org/Rebus/issues/291. 

I've finished (I hope) the core of ```SendAsync``` support (```RebusBus```, configuration etc) but I when started working on the actual transport implementation ```FileSystemMessageQueue```, I had the confusion that I mentioned earlier but, I guess, not properly explained, with the following piece of code:
```if (context.IsTransactional)
{
    context.DoCommit += commitAction;
}
else
{
    commitAction();
}```

In case of ```SendAsync```, ```commitAction``` here should be rather ```Func<Task>``` and awaited later in ```AmbientTransactionContext.Commit```, for instance. So I don't how to resolve this better. Maybe, we should have separate sync and async versions of ```ITransactionContext```, which doesn't sound like a good idea.

Let know if I should try harder explaining what I have in my mind:)

### DIFF
--- a/src/Rebus.Tests/Contracts/ContainerAdapters/TestContainerAdapters.cs
+++ b/src/Rebus.Tests/Contracts/ContainerAdapters/TestContainerAdapters.cs
@@ -1,15 +1,15 @@
-﻿using System;
-using System.Threading;
-using System.Threading.Tasks;
-using NUnit.Framework;
+﻿using NUnit.Framework;
 using Rebus.Configuration;
 using Rebus.Logging;
 using Rebus.Shared;
 using Rebus.Tests.Contracts.ContainerAdapters.Factories;
+using Rebus.Transports.Msmq;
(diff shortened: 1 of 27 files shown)


mookid8000:
Excellent :) I'm pretty busy at the moment, but I'll try to get some time to look at it during the coming week

asgerhallas:
If you send a message async, does that mean that it will be sent before the transaction is committed?

asgerhallas:
@DixonD-git aren't there any implementations of the ISendMessagesAsync interface?

DixonD-git:
@asgerhallas  no, actually that's where I've got stuck. See my first comment, FileSystemMessageQueue was the first transport I tried to adapt to ISendMessagesAsync

mookid8000:
I'm closing this one now because Rebus2 will be async to the core and will have async send by default.

PR #293, opened 2015-02-04
DryIoc

DixonD-git:
Integration for DryIoc container.


mookid8000:
DryIoc container adapter is out in 0.78.1, on NuGet.org within a few minutes - thanks :)

DixonD-git:
It seems that you have put wrong dependency on NuGet - it says  Rebus.DryIoc (≥ 2.0.0-preview181) which should be DryIoc (≥ 2.0.0-preview181)

mookid8000:
oh shit, you're right! I'll push 0.78.2 right away

:smiley:

PR #299, opened 2015-02-18
NewtonSoftJsonMessageSerializer

joshua5822:
Created a NewtonSoftJsonMessageSerializer that can accept
JsonSerializerSettings.
Added AssemblyQualifiedName to Headers.cs, so that the type can be
appended to the message when using NewtonSoftJsonMessageSerializer.

Added UseNewtonSoftJsonSerializer to RebusSerializationConfigurer.cs, so that the new serializer can be used.

### DIFF
--- a/src/Rebus/Configuration/RebusSerializationConfigurer.cs
+++ b/src/Rebus/Configuration/RebusSerializationConfigurer.cs
@@ -1,4 +1,5 @@
 using System.Messaging;
+using Newtonsoft.Json;
 using Rebus.Serialization.Binary;
 using Rebus.Serialization.Json;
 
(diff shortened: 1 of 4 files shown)


seankearon:
I really like this.  Personally, I use JSON to serialise everything and I this will really open up opportunities for libs and tooling to work with messages.  Such as browsing a queue (especially the audit queue) and perhaps even use cases such as modifying a message before returning it to the source queue (dragons notwithstanding!).  Certainly +1 from me! :)

Warning - comments made without fully reviewing the code!

mookid8000:
Excellent :) I'll take a look at it

mookid8000:
I've made a few small changes to make it work properly - since Rebus still supports batching (at least for a while), a transport message _always_ contains an `object[]` which requires full type information to be present in the JSON objects.

Therefore, I've made these changes:
* The header has a ;-separated list of assembly-qualified type names of the logical messages
* Each logical message is serialized to a single-line JSON object in the message body

This way - in cases where you're not using inheritance or doing other funky stuff inside your messages - you can serialize messages as pure JSON objects without .NET type information in the body.

Your contribution is available in 0.80.0 - thanks! :+1:

joshua5822:
No problem

joshua5822:
It works great. The property is being ignored now.

mookid8000:
good to hear :+1:

PR #301, opened 2015-03-06
Fix ASB multiple message receive in the same transaction

AndreaCuneo:
When receiving multiple messages in the same transaction context, the ASB transport would fail because it was doing the following on each receive.

````
if (context[AzureServiceBusMessageBatch] != null)
    throw [...];

context[AzureServiceBusMessageBatch] = new List<Tuple<string, Envelope>>();
````

I triggered this bug while testing the ReturnToSourceQueue.AzureServiceBus tool.

For safeness, I added a check not to fail on commit when no messages haven't been enqueued but the commit has been called. I can remove this bit if you prefer.

I've see no tests section regarding receiving multiple messages in the same transaction. Did I missed them?

### DIFF
--- a/src/Rebus.AzureServiceBus/AzureServiceBusMessageQueue.cs
+++ b/src/Rebus.AzureServiceBus/AzureServiceBusMessageQueue.cs
@@ -196,7 +196,6 @@ namespace Rebus.AzureServiceBus
                         }
 
                         context[AzureServiceBusReceivedMessage] = brokeredMessage;
-                        context[AzureServiceBusMessageBatch] = new List<Tuple<string, Envelope>>();
 
                         // inject method into message context to allow for long-running message handling operations to have their lock renewed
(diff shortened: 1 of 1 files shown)


mookid8000:
Hi @AndreaCuneo , I'm sorry but I had to revert your commit - it made the Azure Service Bus transport tests fail, due to something related to peek lock renewal that did not seem to be handled correctly.

Could you maybe take a look at it again, and then try and run the transport tests that have the `AzureServiceBusMessageQueueFactory` factory on them - thanks :)

AndreaCuneo:
Hi @mookid8000 , I'll check it and let you know. 

Probably this is related to the fact that the transport doesn't expect the need to renew the lock on more than one message. In this case maybe would be better if I just change how the snoop tool works instead of adding the feature of supporting multiple messages in the same transaction for this transport (no one seems to need it).

PR #302, opened 2015-03-18
mongocshardriver 1.10.0 support

CasperTDK:
Any service using the mongocshardriver 1.10.0 (latest stable release) will fail when trying to save sagadata. It only happens at runtime when MongoDbSagaPersister.EnsureResultIsGood is called.

This happened because WriteConcernException is now an internal class and causes everything to fail pretty bad.

I will make a new pull request once mongocshardriver 2.0.0 is released (right now in beta).


CasperTDK:
Also note that the current version of mongocshardriver used in Rebus does not work properly with the WiredTiger engine - as GetCollectionNames will return an empty list in older versions of the driver.
This pull request also solves that issue.

mookid8000:
excellent work :clap: :sunglasses: 

it'll be out in 0.82.0 in a few minutes........

CasperTDK:
Thanks for the merge! Rebus.MongoDb 0.82.0 is still not out though? Only rebus (core) has been updated on nuget.

mookid8000:
ach bummer - NuGet is so flaky, it often times out or something after having pushed half of the packages.... stay tuned, I'll push the rest right now

mookid8000:
the rest of the packages should be on NuGet.org now :)

CasperTDK:
Thanks!

Issue #295, opened 2015-02-06
Allow for competing due timeouts consumers


mookid8000:
has already been implemented with SQL Server since 0.79.0, using a row-level update lock and readpast

something similar ought to work with Postgres

MongoDB should use findAndModify to obtain a "lease" on the due timeout

Not sure about RavenDB, though - I wonder if it's even possible

CasperTDK:
Could you elaborate what you mean by 'lease' regarding mongoDB?
I can't see a way to avoid removing the timeout documents untill 'MarkAsProcessed' unless you do something like http://docs.mongodb.org/manual/tutorial/perform-two-phase-commits/

I've tried submitting a solution. What do you think?

CasperTDK:
I guess you could expand the update statement to add a timestamp = now+ 100 ms - to avoid multiple consumers for a message
This would cause issues if the process hangs and not just crashes though

mookid8000:
A common way to implement "receiving" something from MongoDB as if it was a message queue, is to use FindAndModify where you update the found document by setting a time stamp on it, which would function as a lease for, say, 1 minute or something like that.

The receiving process will then have one full minute to turn it into a timeout reply and then delete the document to complete the transaction so to speak. 



Sendt fra min iPhone

CasperTDK:
Ah, that is what my solution did (see commits in this thread). I will make
a pull request soon

On Fri, Mar 20, 2015 at 4:26 PM, Mogens Heller Grabe <
notifications@github.com> wrote:

> A common way to implement "receiving" something from MongoDB as if it was
> a message queue, is to use FindAndModify where you update the found
> document by setting a time stamp on it, which would function as a lease
> for, say, 1 minute or something like that.
>
> The receiving process will then have one full minute to turn it into a
> timeout reply and then delete the document to complete the transaction so
> to speak.
>
>
>
> Sendt fra min iPhone
>
> > Den 20/03/2015 kl. 07.12 skrev Casper Thygesen <notifications@github.com
> >:
> >
> > I guess you could expand the update statement to add a timestamp = now+
> 100 ms - to avoid multiple consumers for a message
> > This would cause issues if the process hangs and not just crashes though
> >
> > —
> > Reply to this email directly or view it on GitHub.
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rebus-org/Rebus/issues/295#issuecomment-84048905>.
>



-- 
Med venlig hilsen
Casper

[...]

PR #304, opened 2015-03-20
Feature competing timeout consumers #295 (mongoDB)

CasperTDK:


mookid8000:
Not sure I follow your code. Before I read it again, I'll briefly explain how the thing with the lease could work.

Due timeouts should be retrieved with a find-and-modify query that queries for

    timeoutIsDue && (leaseHasNotBeenSet || leaseHasExpired)

The `leaseHasNotBeenSet || leaseHasExpired` part can be made easy if the lease defaults to `DateTime.Min`, because then an unset lease and an expired lease will look the same. Thus the query can be reduced to

    timeoutIsDue && leaseHasExpired

which translates to something like the following query and update:

    var now = RebusTimeMachine.Now();
    var criteria = Query.And(Query.LTE(TimeProperty, now), Query.LTE(LeaseProperty, now));

    var leaseExpirationTime = now + TimeSpan.FromMinutes(1);
    var update = Update.Set(LeaseProperty, leaseExpirationTime);
    
As you've figured out, MongoDB's `findAndModify` returns only 1 matching document - therefore, some kind of iteration is necessary, fetching up to e.g. 100 timeouts at a time:

    var dueTimeouts = Enumerable.Range(0, 100)
        .Select(n => GetNextTimeout(criteria, update))
        .TakeWhile(timeout => timeout != null)
        .ToList();

Does it make sense?

CasperTDK:
This is what the code should already do :)

mookid8000:
...but what's the deal with the ultra-short 500 ms lease and the `boundTo` construct?

CasperTDK:
The 500ms is the lease - do you think it will ever take more than 500ms to retrieve and send timeout replies found?
The boundTo ensures that we can iterate 'FindAndModify' and get all documents - and not the same document on every call.

mookid8000:
I'm thinking that retrieving up to 100 timeouts could take a few seconds - the lease must be long enough that even e.g. an Azure-deployed web site in West Europe can pull timeouts from a MongoLab-hosted MongoDB in West US and convert those into timeout replies and delete the timeout documents in time.

Also, the [lease must function as a time-limited lock](http://en.wikipedia.org/wiki/Lease_%28computer_science%29) - which is exactly what it does when you use `findAndModify`: the timeout document will simply not turn up in results until the lease expires....

mookid8000:
PS: I'm not saying that your solution does not work - I'm just hinting at a simpler solution that is easier to understand :)

mookid8000:
Rebus2 has a timeout manager for MongoDB that works with competing consumers

Issue #306, opened 2015-04-20
FileSystem transport - file locking


mgibas:
I'm using filesystem transport and have a problem problem with locking files:

```csharp
The process cannot access the file 'XXXXXXXX.rebusmessage' because it is being used by another process.
```

It's situation where two processes want to send message to same queue. For that we have this piece of code:

FileSystemMessageQueue.cs
```csharp
public void Send(string destinationQueueName, TransportMessageToSend message, ITransactionContext context)
{
    EnsureQueueInitialized(destinationQueueName);

    var destinationDirectory = GetDirectoryForQueueNamed(destinationQueueName);

    var serializedMessage = Serialize(message);
    var fileName = GetNextFileName();
    var fullPath = Path.Combine(destinationDirectory, fileName);

    Action commitAction = () => File.WriteAllText(fullPath, serializedMessage, FavoriteEncoding);

    if (context.IsTransactional)
    {
        context.DoCommit += commitAction;
    }
    else
    {
        commitAction();
    }
}
```

in this line we are interlocking msg number, but thats not enough if we talk about multiple processes
```csharp
var fileName = GetNextFileName();
```

Any ideas how to overcome this ? Maybe some king of name mangling (with i.e. guid ?)

mgibas:
After quick "live" test it seams that switching from:

```csharp
string GetNextFileName()
{
    return string.Format("{0:yyyy_MM_dd_HH_mm_ss}_{1:0000000}.rebusmessage",
        DateTime.UtcNow, Interlocked.Increment(ref incrementingCounter));
}
```
to
```csharp
string GetNextFileName()
{
    return string.Format("{0}_{1:0000000}.rebusmessage",
        Guid.NewGuid(), Interlocked.Increment(ref incrementingCounter));
}
```
works as expected. Let me know if this change makes any sens - I will create PR :)

asgerhallas:
Yes, I believe the easiest way is to concat some uniqueness at the end of the filename. That will make the order of concurrent messages random, but that - hopefully - shouldn't matter to the application. It could be suffixed with a guid or maybe the bus instance name.

asgerhallas:
Oh, I was too slow there... I believe you should let the guid be suffixed to keep the order. I'm not sure, but I guess the messages are read in order by name.

mgibas:
Ya, good point:
```csharp
var fileNames = Directory.GetFiles(GetDirectoryForQueueNamed(inputQueue), "*.rebusmessage")
[...]

PR #307, opened 2015-04-20
FileSystem transport - file locking

mgibas:
Fix for #306 

Suffixing message filename with guid.


mookid8000:
cool, thanks! - it's out in 0.83.0 in a few minutes :)

PR #321, opened 2015-08-11
Switch SQL Server implementations to use IDbconnectionProvider

MrMDavidson:
Currently `SqlServerTransport` makes use of an `IDbConnectionProvider` however `SqlServerSagaStorage`, `SqlServerSubscriptionStorage` and `SqlServerTimeoutManager` all make use of the concrete `DbConnectionProvider`. This quick patch relaxes this so that they all make use of `IDbConnectionProvider` instead to allow consumer to provide their own implementations of `IDbConnectionProvider`.


mookid8000:
:+1: merged in 0.91.0 which will be on NuGet shortly

MrMDavidson:
Excellent!

Issue #187, opened 2013-08-16
Optimizing SQL Server transport even further


improvedk:
Just hit me you might actually be able to optimize the transport even further...

Right now you do something like this:
* Begin tran
* Select & lock available message from queue [DB roundtrip]
* Process message
* Delete locked message from queue [DB roundtrip]
* Commit tran [DB roundtrip]

Instead you should be able to do:
* Begin tran
* Delete available message from queue - using the OUTPUT clause you can get the deleted row back to the client [DB roundtrip]
* Process message
* Commit tran [DB roundtrip]

That way you only have *one* DML statement instead of selecting & deleting separately. In your non-transactional mode you could simply delete the message outright and just hope the client processes it, though that probably won't satisfy the once-or-more constraint as it'll be zero-or-one. But then again, as most will probably be transactional, you could have the optimized version only for the transactional version.

mookid8000:
I'm parking this one for now :)

PR #324, opened 2015-08-13
Use a DELETE with OUTPUT in SqlServerTransport

MrMDavidson:
I saw you had Issue #187 from a while back suggesting you switch to using `DELETE` with `OUTPUT`. I, too, thought that this was a bit weird and non-performant. In the past I've used a CTE with `DELETE` and an `OUTPUT` clause for a similar implementation and it's worked flawlessly. The performance of the two individual `SELECT` and `DELETE` commands is made even worse in an environment like Azure.

I whipped up a small test program that sends a message to the bus which has a simple handler that prints a message every 100 times it handles the message. I then ran multiple copies of this (Roles column below) and add workers (Workers column below) and logged the performance of processing 10,000 messages. The timings here are from the point of initially queuing (done with serial a `bus.Send(...).Wait()`) until a `SELECT COUNT(1) FROM ...  WHERE Recipient = ...` returns 0 results.

I only ran these once per iteration but it's pretty clear `DELETE` with `OUTPUT` is much quicker than `SELECT` with a separate `DELETE`. These timings are running on a local SQL Server instance on an Intel i7-4790 CPU with 16GB RAM.

Messages | Workers | Roles | DELETE-OUTPUT Time (s) | SELECT+DELETE Time (s) | DELETE-OUTPUT Messages / Second | SELECT+DELETE Messages / Second
-------- | ------- | ----- | ---------------------- | ---------------------- | --------------------------------| -------------------------------
10000 | 1 | 1 | 11.5070503 | 102.501546 | 869.0324401 | 97.55950413
10000 | 2 | 1 | 9.2873886 | 92.2290834 | 1076.728931 | 108.4256683
10000 | 3 | 1 | 9.8928232 | 72.3371684 | 1010.833793 | 138.2415185
10000 | 4 | 1 | 8.2214449 | 80.23932652 | 1216.331207 | 124.6271677
10000 | 5 | 1 | 9.4639883 | 54.1196487 | 1056.636978 | 184.7757744
10000 | 6 | 1 | 10.9329288 | 70.8496198 | 914.6679891 | 141.1440178
10000 | 7 | 1 | 8.60002424 | 58.4448878 | 1162.78742 | 171.1013636
10000 | 8 | 1 | 8.5476005 | 57.3613015 | 1169.918973 | 174.3335618
10000 | 1 | 2 | 9.1795095 | 89.6733901 | 1089.382826 | 111.5158018
10000 | 2 | 2 | 9.7746501 | 80.3168243 | 1023.054523 | 124.5069148
10000 | 4 | 2 | 8.4508077 | 62.9816794 | 1183.318844 | 158.7763314
10000 | 8 | 2 | 8.655181 | 68.2099974 | 1155.37734 | 146.6060751
10000 | 1 | 3 | 9.2892266 | 81.3148503 | 1076.515886 | 122.9787666
10000 | 2 | 3 | 10.3012478 | 79.4531262 | 970.7561835 | 125.8603717
10000 | 4 | 3 | 9.0368907 | 63.5257153 | 1106.575296 | 157.4165667
10000 | 8 | 3 | 9.7584354 | 60.7622159 | 1024.75444 | 164.5759598
10000 | 1 | 4 | 9.023949 | 84.8348466 | 1108.162291 | 117.8760898
10000 | 2 | 4 | 8.411234 | 73.9413074 | 1188.886197 | 135.2424017
10000 | 4 | 4 | 8.7418844 | 59.2433966 | 1143.918124 | 168.7951835
10000 | 8 | 4 | 9.5626814 | 51.5188278 | 1045.731797 | 194.1037952

### DIFF
--- a/Rebus/Transport/SqlServer/SqlServerTransport.cs
+++ b/Rebus/Transport/SqlServer/SqlServerTransport.cs
@@ -218,23 +218,28 @@ VALUES
 
                 using (var selectCommand = connection.CreateCommand())
                 {
-                    selectCommand.CommandText =
-                        string.Format(@"
-SELECT TOP 1
-    [id],
-    [headers],
-    [body]
-FROM [{0}] 
-WITH (UPDLOCK, READPAST, ROWLOCK)
-WHERE 
-    [recipient] = @recipient 
-    AND [visible] < getdate()
-    AND [expiration] > getdate()
-ORDER BY 
-    [priority] ASC, 
-    [id] asc
+                    selectCommand.CommandText = string.Format(@"
+	SET NOCOUNT ON
 
-", _tableName);
+	;WITH TopCTE AS (
+		SELECT	TOP 1
+				[id],
+				[headers],
+				[body]
+		FROM	{0} M WITH (ROWLOCK, READPAST)
+		WHERE	M.[recipient] = @recipient
+		AND		M.[visible] < getdate()
+		AND		M.[expiration] > getdate()
+		ORDER
+		BY		[priority] ASC,
+				[id] ASC
+	)
+	DELETE	FROM TopCTE
+	OUTPUT	deleted.[id] as [id],
+			deleted.[headers] as [headers],
+			deleted.[body] as [body]
+						
+						", _tableName);
 
                     selectCommand.Parameters.Add("recipient", SqlDbType.NVarChar, RecipientColumnSize).Value = _inputQueueName;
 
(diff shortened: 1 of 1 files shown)


mookid8000:
awesome work! it's out in 0.92.3 in a few minutes :)

asgerhallas:
Very very nice! :)

MrMDavidson:
Sweet!

Were you able to compare the numbers on your setup? In the linked issue you were getting worse numbers compared to the original select+delete method.

mookid8000:
To tell you the truth, I blindly trusted your numbers (but I did run the tests though :))

I was also wondering about the bad `SqlTransport` performance in Rebus2, because in Rebus1 IIRC I got around 1000 msg/s too, and a little bit less with the DELETE OUTPUT version.

All (the few I've done) of my performance tests of the old SQL transport were done on my local box, so they have been artificial in the sense that the db communication was done with shared memory protocol, and thus they were unlikely to reflect the cost of issuing two commands instead of one.

MrMDavidson:
I haven't compared the old and current implementations of the transport but with the original Rebus 2 version we were seeing very little load on our DB servers but message rates of ~20 (Azure, S3) and ~60 (Local, i7) messages per second for a simple message handler (ie. Not the one used in this table).

Tomorrow (Australia time) I'll deploy to Azure with 0.92.3+ and see what sort of rates we get for that same handler.

mookid8000:
Cool - looking forward to hearing about your findings :)

PR #337, opened 2015-09-02
Add Type to SqlServerTransport

MrMDavidson:
Two fairly minor changes;
 1. Add Type to the message table. This is extracted from the headers and persisted here. It's not used by Rebus but is very helpful for queue introspection and monitoring
 2. Along the same lines; re-made `HeaderSerializer` public (we again use it to extract some properties from the messages as part of our dashboarding). I imagine the long term solution would have some sort of generic serializer instance you can retrieve through the API? But I'm not sure how that would work with other transports.

### DIFF
--- a/Rebus/Transport/SqlServer/SqlServerTransport.cs
+++ b/Rebus/Transport/SqlServer/SqlServerTransport.cs
@@ -130,6 +130,7 @@ CREATE TABLE [dbo].[{0}]
 (
 	[id] [bigint] IDENTITY(1,1) NOT NULL,
 	[recipient] [nvarchar](200) NOT NULL,
+	[type] [nvarchar](max) NULL,
 	[priority] [int] NOT NULL,
     [expiration] [datetime2] NOT NULL,
     [visible] [datetime2] NOT NULL,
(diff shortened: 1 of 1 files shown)


mookid8000:
Cool, thanks - but it makes me think about why message type should be treated as more important than other headers - e.g. correlation ID could may be be interesting too, and maybe other headers that you could have added yourself.

If it doesn't cause too much of a performance hit, maybe the headers could be serialized into an [XML column](https://msdn.microsoft.com/en-us/library/hh403385.aspx)? This way, all headers would be accessible "from the outside" (which they actually should be, at least according to [the description of Rebus' wire-level format](https://github.com/rebus-org/Rebus/wiki/Wire-level-format-of-messages#message-headers).

MrMDavidson:
Yeah - that's a valid actually a very valid point. The type shouldn't be any more special than anything else. In our case it's just so we can get a simple depth-per-message count.

I like the idea of the headers being visible to the outside and the XML column probably achieves that. I've not had a great deal of experience with them in recent times so I'm not sure how performant they are. I'll have a think about this.

PR #351, opened 2015-10-06
Update ASB transports to use streams with brokered message body

Meyce:
Streams provide better interop with other languages/platforms due to the way the body of the brokered message is serialized.  This does represent a breaking change in the event messages were sent/published with a previous version of the transport.  If this is a problem an option could probably be added to the AzureServiceBusTransportSettings class to opt-in to the stream approach.

### DIFF
--- a/Rebus.AzureServiceBus/AzureServiceBusTransport.cs
+++ b/Rebus.AzureServiceBus/AzureServiceBusTransport.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Concurrent;
+using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.ServiceBus;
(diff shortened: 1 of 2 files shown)


mookid8000:
excellent work, thanks :+1: 

It'll be out in 0.99.0 as soon as the build server has finished crunching

Meyce:
No problem. Glad to help.

On Tuesday, October 6, 2015, Mogens Heller Grabe <notifications@github.com>
wrote:

> excellent work, thanks [image: :+1:]
>
> It'll be out in 0.99.0 as soon as the build server has finished crunching
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rebus-org/Rebus/pull/351#issuecomment-146004352>.
>


-- 
Sent from Gmail Mobile

PR #356, opened 2015-10-30
Fix for using remote MSMQ

akoso:
I had an issue with using an MSMQ on a remote machine, because when getting the queue I believe always a local path is used which is then checked for existence and whether it is transactional.

I've changed how the MSMQ path is resolved based on whether it's local or remote, so we get the full path for a remote queue. For a local queue everything is the same as before but for a remote one we don't try to do the existence and transctionality checks because they don't work (they throw: `System.Messaging.MessageQueueException: The specified format name does not support the requested operation. For example, a direct queue format name cannot be deleted.`)

### DIFF
--- a/Rebus/Transport/Msmq/MsmqTransport.cs
+++ b/Rebus/Transport/Msmq/MsmqTransport.cs
@@ -306,11 +306,19 @@ namespace Rebus.Transport.Msmq
             {
                 if (_inputQueue != null) return _inputQueue;
 
-                var inputQueuePath = MsmqUtil.GetPath(_inputQueueName);
+                string inputQueuePath;
 
-                MsmqUtil.EnsureQueueExists(inputQueuePath, _log);
+                if (MsmqUtil.IsLocal(_inputQueueName))
+                {
+                    inputQueuePath = MsmqUtil.GetPath(_inputQueueName);
 
-                MsmqUtil.EnsureMessageQueueIsTransactional(inputQueuePath);
+                    MsmqUtil.EnsureQueueExists(inputQueuePath, _log);
+                    MsmqUtil.EnsureMessageQueueIsTransactional(inputQueuePath);
+                }
+                else
+                {
+                    inputQueuePath = MsmqUtil.GetFullPath(_inputQueueName);
+                }
 
                 _inputQueue = new MessageQueue(inputQueuePath, QueueAccessMode.SendAndReceive)
                 {


mookid8000:
Hi @akoso , thanks for your PR!

With old Rebus, I actively prohibited a configuration that would result in receiving from a remote MSMQ queue, [because it is inefficient and is not the optimal way of using MSMQ](https://msdn.microsoft.com/en-us/library/ms699854(v=vs.85).aspx).

I do understand though that it can function as a "poor man's scaling option", because MSMQ _can_ actually work in a competing consumers setup - it's just that it works up to 2-3 consumers, and after that the overhead becomes too big, and you will no longer increase throughput by adding consumers.

I actually intented to prohibit remote receive on Rebus 2 too, but since we're at it, let's talk about whether it should be possible :)

I'm curios: why you want to use a remote MSMQ queue as your input queue?

akoso:
Hi @mookid8000 ,

First of all thanks for the really fast reply, and let me note that I'm a newbie with both MSMQ and Rebus (great work there, keep it up :)), so likely I'm trying to solve the wrong problem.

I understand from your comment that using MSMQ this way is probably not a good idea, so it would be good to just throw an exception there for remote queues that this is explicitly not supported.

However maybe you could guide me in the right direction. What I have is a service and a web frontend (MVC) which are sending messages to each other (in both directions) via Rebus and MSMQ. This works all good while they are both running on the same machine using a single queue, however what is not really clear for me is what to do when they are deployed to different servers. Here I just tried to use the same queue on both ends, with which it became a remote queue on one end (hence the issues and the PR). I would really appreciate if you could just point out whether I'm trying to do something wrong and possibly give a few tips. Maybe I shouldn't be even using MSMQ, but rather Azure Service Bus, or maybe something else?

mookid8000:
The recommended solution in that case would be for each endpoint to receive their messages from a local queue, and then just send to the other endpoint's remote queue.

E.g. if MACHINE1 has a Rebus endpoint with local input queue "web", this endpoint can be addressed from anywhere by using `web@MACHINE1` as the queue name.

You will usually want to set this up in the endpoint's endpoint mappings:

    Configure.With(...)
        .(...)
        .Routing(r => r.TypeBased()
            .MapAssemblyOf<WebMessage>("web@MACHINE1")
            .MapAssemblyOf<ServiceMessage("service@MACHINE2"))
        .(...)

I hope that makes sense :smiley:

akoso:
Hi @mookid8000 ,

Thank you, that makes a lot of sense. 

And as it turns out I was trying to solve the wrong problem, so I'll just close this PR.

mookid8000:
Excellent :+1: I'm looking forward to your next PR :wink:

PR #361, opened 2015-11-10
Removed some extra " ; " tags + removed some default values from the method itself

SvenVandenbrande:
Clean-Up....


mookid8000:
thanks :+1:

PR #364, opened 2015-11-17
Master

puzsol:
Adding LightInject to the family of IoC frameworks


mookid8000:
LightInject container adapter is out [on NuGet.org](https://www.nuget.org/packages/Rebus.LightInject/) with Rebus 0.99.12

Thanks :+1: :smiley:

PR #371, opened 2015-12-01
Convention based subscriptions.

gertjvr:
Added Subscribe / Unsubscribe overloads to allow for convention based subscriptions.

Hit a situation today where I wanted to subscribe all event based messages that were being handled, had to use refection to achieve this.

```
foreach (var eventType in eventTypes)
{
    var method = bus.GetType().GetMethod("Subscribe");
    var generic = method.MakeGenericMethod(messageType);
    var task = (Task)generic.Invoke(bus, null);
    task.Wait();
}
```

This will allow for easier convention based subscriptions.


mookid8000:
excellent, thanks :+1: :smile:

PR #375, opened 2015-12-09
Added test categories to allow us to ignore them.

gertjvr:
This is allowing us to ignore some of the integration tests that we don't use.


mookid8000:
:+1:

PR #377, opened 2015-12-10
Found a nice bug, with the AmazonSQS SendMessageBatchAsync

gertjvr:
SendMessageBatchAsync doesn't throw any exceptions, you need to look at the failed collection to see the failures, not sure what would happen by throwing an aggregate exception or what the recovery process would be.

I guess this is to start a discussion how to fix this issue.

### DIFF
--- a/Rebus.AmazonSQS/AmazonSQSTransport.cs
+++ b/Rebus.AmazonSQS/AmazonSQSTransport.cs
@@ -222,7 +222,13 @@ namespace Rebus.AmazonSQS
 
                         var request = new SendMessageBatchRequest(destinationUrl, entries);
 
-                        await client.SendMessageBatchAsync(request);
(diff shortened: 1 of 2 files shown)


mookid8000:
:scream: :+1: :hand: 

@hdrachmann damn! I thought we fixed that :laughing:

mookid8000:
when the build server has finished munching through the tests I'll release your fix in 0.99.20

mookid8000:
0.99.20 is on NuGet.org now.... had some trouble with the build server today, but it's fine now :grimacing:

gertjvr:
Thanks will have a look today, unrelated, how do you prefer discussing features? GH issue or stackoverflow?

mookid8000:
Features and issues (as in problems that need fixing) on GitHub, issues (as in questions regarding usage) on StackOverflow. 



Sendt fra min iPhone

PR #387, opened 2016-01-20
Change optionsconfigurer extension methods to return void 

xenoputtss:
I updated all extension methods for OptionsConfigurer that returned OptionsConfigurer to return void.  I updated corresponding references in code.

All Tests pass, I was unable to run integration.persistence+transports tests because I do not have the appropriate setups for those.


mookid8000:
excellent - thanks :+1: 

the API changes will be out in 0.99.25

PR #389, opened 2016-01-21
Populate headers from the sqs message if not present.

gertjvr:
We have a setup where messages are push to sqs queue from sns-topic so we dont control how the message was formed, these changes allows to shape the sns message to something rebus would understand. 

It does contain a breaking change in `GetBody` and `GetBodyBytes` UTF8 encodes the body rather than using base64

### DIFF
--- a/Rebus.AmazonSQS/AmazonSQSTransport.cs
+++ b/Rebus.AmazonSQS/AmazonSQSTransport.cs
@@ -380,17 +380,32 @@ namespace Rebus.AmazonSQS
         {
             var headers = message.MessageAttributes.ToDictionary(kv => kv.Key, kv => kv.Value.StringValue);
 
+            if (!headers.ContainsKey(Headers.ContentType))
+                headers[Headers.ContentType] = "application/json;charset=utf-8";
+
+            if (!headers.ContainsKey(Headers.ReturnAddress))
+                headers[Headers.ReturnAddress] = Address;
+
+            if (!headers.ContainsKey(Headers.MessageId))
+                headers[Headers.MessageId] = message.MessageId;
+
+            if (!headers.ContainsKey(Headers.CorrelationId))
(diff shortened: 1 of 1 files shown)


Rebus.AmazonSQS/AmazonSQSTransport.cs
```diff
@@ -383,17 +383,32 @@ private TransportMessage GetTransportMessage(Message message)
...
+            if (!headers.ContainsKey(Headers.Intent))
+                headers[Headers.Intent] = Headers.IntentOptions.PublishSubscribe;
+
             return new TransportMessage(headers, GetBodyBytes(message.Body));
 
         }
         private string GetBody(byte[] bodyBytes)
         {
-            return Convert.ToBase64String(bodyBytes);
+            return System.Text.Encoding.UTF8.GetString(bodyBytes);
```
mookid8000: Sorry, but this will not work in the general case.... it is not guaranteed that the configured serializer is a JSON serializer, and even if it was a JSON serializer the transport should not assume UTF8 encoding

Is there no way for Amazon SQS to send raw bytes?

gertjvr:
>Is there no way for Amazon SQS to send raw bytes? 

Looking at the Amazon SQS api `SendMessageBatchRequestEntry` MessageBody only accepts string.

gertjvr:
When I was working on this thought that dynamic content negotiation would be a nice feature.

ie inspect content-type message header and pick the right deserializer  :smile:

mookid8000:
You can easily implement varying deserialization strategies based on headers by creating your own serializer decorator that wraps multiple serializers. 

This would work well and simple for incoming messages, but you would need to figure out some strategy for picking the serializer to use for outgoing messages.

gertjvr:
I see the error by selecting UTF8 encoding, the only general way forward would be to inspect the message and test if its base64 encoded before decoding it. Will try and resolve that this week

PR #391, opened 2016-01-21
Some enhancements to Rebus.Autofac

gertjvr:
- Allows you to register `AutofacContainerAdapter` with a `LifetimeScope`
- Added a key to allow to resolve handlers based on the provided key.
- If you register using a `LifetimeScope` it will not update the container, you have to call builder.RegisterRebus(). (updating an autofac container should be avoided if possible)

### DIFF
--- a/Rebus.Autofac/AutofacContainerAdapter.cs
+++ b/Rebus.Autofac/AutofacContainerAdapter.cs
@@ -19,13 +19,25 @@ namespace Rebus.Autofac
     public class AutofacContainerAdapter : IContainerAdapter
     {
         readonly IContainer _container;
+        private readonly ILifetimeScope _scope;
+        private readonly object _key;
 
         /// <summary>
         /// Constructs the adapter, using the specified container
         /// </summary>
-        public AutofacContainerAdapter(IContainer container)
+        public AutofacContainerAdapter(IContainer container, object key = null)
         {
             _container = container;
+            _key = key;
+        }
+
+        /// <summary>
+        /// Constructs the adapter, using the specified lifetime scope
+        /// </summary>
+        public AutofacContainerAdapter(ILifetimeScope scope, object key = null)
+        {
+            _scope = scope;
+            _key = key;
         }
 
         /// <summary>
@@ -36,7 +48,7 @@ namespace Rebus.Autofac
             var lifetimeScope = transactionContext
                 .GetOrAdd("current-autofac-lifetime-scope", () =>
                 {
-                    var scope = _container.BeginLifetimeScope();
+                    var scope = (_container ?? _scope).BeginLifetimeScope();
 
                     transactionContext.OnDisposed(() => scope.Dispose());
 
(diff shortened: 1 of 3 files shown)


gertjvr:
Example of a using a autofac module to register a IBus 

```
public class RabbitMqRebusModule : Module
    {
        protected override void Load(ContainerBuilder builder)
        {
            builder.Register(c => Configure.With(new AutofacContainerAdapter(c.Resolve<ILifetimeScope>()))
                    .Logging(l => l.Serilog(c.Resolve<ILogger>()))
                    .Transport(t => t.UseRabbitMq("amqp://localhost", "input"))
                    .Start())
                .As<IBus>()
                .SingleInstance();

            builder.RegisterRebus();
        }
    }
```

mookid8000:
Hi @gertjvr , nice to see your level af activity!

I'm not an expert on Autofac - can you explain to me what is achieved by the `key` parameter?

gertjvr:
Key is two fold, if you have multiple rebus instance registered with your container you can resolve a specific instance by using the key. Secondly you can if you used scanning registration for you handlers it will only resolve handlers that has been attributed with that key attribute.

gertjvr:
[LinqPad](http://share.linqpad.net/qdk2tn.linq) example

```
const string SqsKey = "Sqs";
const string RabbitMqKey = "RabbitMq";

void Main()
{
	var builder = new ContainerBuilder();
	
	builder
		.RegisterType<SqsBus>()
		.Keyed(SqsKey, typeof(IBus))
		.InstancePerLifetimeScope();

	builder
		.RegisterType<RabbitMqBus>()
		.Keyed(RabbitMqKey, typeof(IBus))
		.As<IBus>()
		.InstancePerLifetimeScope();
	
	var container = builder.Build();
	
	//Will resolve sqs bus instance
	var bus1 = container.ResolveKeyed<IBus>(SqsKey);
	bus1.Send("Hello");

	//Will rabbitmq bus instance
	var bus2 = container.ResolveKeyed<IBus>(RabbitMqKey);
	bus2.Send("Hello");
	
	//Will only resolve rabbitmq bus instance
	var instances = container.Resolve<IEnumerable<IBus>>();
	foreach (var instance in instances)
	{
		instance.Send("Hello");	
	}
}
```

mookid8000:
It sounds to me like this solves a problem that you should not have in the first place. Why would you register multiple buses in the same container?

gertjvr:
I can think of many reasons... in one services I have two busses on to receive message from sqs and republish it on our internal rabbit cluster.

mookid8000:
Right :) 

I generally recommend people to only configure one bus per container instance.

This means that if your process hosts more buses, it will host several containers - which is a the right model, because a container is supposed to function as a logical application of sorts, which fits nicely with being an autonomously running Rebus endpoint.

I have seen so many examples of people attempting to use the same container instance to contain both an ASP.NET Web API and a Rebus bus, but this is _painful_ because people always run into issues around component lifestyles etc.

If you use your Rebus endpoint to be a bridge between to messaging networks (which is what you do when you e.g. receive messages from SQS and publish to RabbitMQ), I would argue that your application is a SQS-enabled Rebus endpoint, which happens to have a RabbitMQ client that it can use.

The fact that this client is a Rebus bus, should not pollute the fact that the endpoint is an SQS-enabled Rebus endpoint - in fact, its RabbitMQ-enabled Rebus bus will most likely be a one-way client.

To avoid injecting `IBus` and being in doubt whether its the bus server (in your case the one that receives messages from SQS) or the bus client (the one that publishes to RabbitMQ), I always recommend people to wrap the client bus instance in something, so that the container need only know one `IBus`.

In your case it could probably look somewhat like this:

	public class RabbitMqBridge : IDisposable
	{
		readonly IBus _bus;

		public RabbitMqBridge()
		{
			_bus = Configure.With(new BuiltinHandlerActivator())
				.Transport(t => t.UseRabbitMqAsOneWayClient("amqp://blabla"))
				.Start();
		}

		public async Task Publish(object message)
		{
			await _bus.Publish(message);
		}
	}

gertjvr:
I generally see a `container` as a `composition root` and only have one container per application it make it easier to verify that everything has been correctly registered and makes `pure construction injections` possible. Having multiple container just move the problem of choosing the wrong bus to choosing the wrong container.

By I have been spoiled by autofac lifetime scope and life cycle management.

And as you said people can get it wrong by choosing the wrong bus / container, this is what the implementation provides is away to precisely select which instance of the bus and handlers you are targeting.

This implementation doesn't change the old package by rather just enhances the autofac implementation giving the developer more options.

mookid8000:
While you are right that this addition does not break existing code, it is not an "enhancement" in my opinion. It makes the Autofac container adapter more capable, sure, but the added feature makes the API of the container adapter harder to understand.

My opinion still is that this PR solves a problem that you should not have in the first place. And the solution is one that allows for keeping something implicit (bus selection), which IMO should have been made explicit (one bus is the host, the other one is a client to another bus network).

mookid8000:
But thanks again for proposing this change :) you should of course just keep your version of the `Rebus.Autofac` package, it's just not something I feel I can bring into Rebus because of the reasons I wrote above.

Thanks again :smiley:

PR #392, opened 2016-01-22
Added outgoing correlation id to Serilog log context.

gertjvr:
Found an issue with serilog / seq couldn't see a correlation id on the initial sent message, everything after the message was received had the correct correlation id.

I have ran all the local tests and all the tests passed.

### DIFF
--- a/Rebus.Serilog.Tests/Properties/AssemblyInfo.cs
+++ b/Rebus.Serilog.Tests/Properties/AssemblyInfo.cs
@@ -1,0 +1,36 @@
+﻿using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("Rebus.Serilog.Tests")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("")]
+[assembly: AssemblyProduct("Rebus.Serilog.Tests")]
+[assembly: AssemblyCopyright("Copyright ©  2016")]
+[assembly: AssemblyTrademark("")]
+[assembly: AssemblyCulture("")]
+
+// Setting ComVisible to false makes the types in this assembly not visible 
+// to COM components.  If you need to access a type in this assembly from 
+// COM, set the ComVisible attribute to true on that type.
+[assembly: ComVisible(false)]
+
+// The following GUID is for the ID of the typelib if this project is exposed to COM
+[assembly: Guid("48fa32d8-bdf1-492d-a9f8-16c3cbae39e3")]
+
+// Version information for an assembly consists of the following four values:
+//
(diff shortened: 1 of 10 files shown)


mookid8000:
I'm sorry but I don't think I understand the problem.... when you say "couldn't see a correlation id on the initial sent message" it sounds to me like you expect a sender to log the correlation ID of _the sent message_...? I'm wondering which log statements it is that you think were missing the correlation ID?

gertjvr:
The issue is the RebusCorrelationIdEnricher only enriches incoming messages. 
the initial outgoing message doesn't have a correlation id, even though I added it the the message optional headers.

gertjvr:
its friday and end of my day, so my english is failing me :disappointed: here an image 
was expecting to see a `CorrelationId {Random-Guid}`
screen shot 2016-01-22 at 4 53 01 pm

mookid8000:
oh, but that is totally the way it should work!

when you are logging things, and your log events have a correlation ID, it is basically like saying "this thing that I'm logging here happened as a consequence of `233343some-correlation-id`, which defaults to be the message ID of the first message that was sent

when you SEND a message, you either do that in the context of a correlation ID (because you're currently handling a message that has a correlation ID), or you don't.... therefore I think it would not be right to log the correlation ID of an outgoing message

gertjvr:
Ok but the i would expect to see the message id. 

There is also a case where the sending of a message is the result of another non bus related request that has a correlation id and would like to carry that id.

mookid8000:
> Ok but the i would expect to see the message id.

You could easily add a transport decorator or an outgoing pipeline step that logs the outgoing message just the way you like it :)

The reason I'm reluctant to do this, is that the sending of a message is not done in the context of the sent message. The sending of the message is either done a) in the context of a handled message (Rebus can log that), or b) in the context of something else (Rebus has no way of knowing how to log that).

> There is also a case where the sending of a message is the result of another non bus related request that has a correlation id and would like to carry that id.

I would say that is is clearly a case of (b) above - and Rebus should IMO not know or care about how to log that.

I hope this makes sense :) you should know that I really appreciate your activity, but I also want to make sure that Rebus does not lose its focus over time.

gertjvr:
I appreciate that you want to keep Rebus focused.

Agree that in terms of rebus its doesn't care much how its been called or what the origin of the message might be. But in a larger picture where the bus is just one component, carrying some sort of correlation id between each parts help diagnose issue early.

Have been using structured logging (serilog / seq) over 2 years and looking back not sure how we survived without it is in dispensable developing distributed systems.

@nblumhardt created a very nice [blog post](http://nblumhardt.com/2015/01/designing-log-events-for-effective-correlation/) explaining what I am trying to achieve.

mookid8000:
> But in a larger picture where the bus is just one component, carrying some sort of correlation id between each parts help diagnose issue early.

I am not at all arguing against this.

In fact, I encourage you to check out the `LogContext` way of logging, as described on [the link you included above](http://nblumhardt.com/2015/01/designing-log-events-for-effective-correlation/) - it could be used to provide the context you want even when you're sending Rebus messages from outside of a Rebus message handler.

PR #394, opened 2016-01-22
Rabbitmqtransportchange

mayermd:
I needed the ability to create transient queues for a use case I have.  This code changes allows the autoDelete, durable and exclusive parameters to be specified for the QueueDeclare call in the RabbitMqTransport class.

### DIFF
--- a/Rebus.RabbitMq/RabbitMqConfigurationExtensions.cs
+++ b/Rebus.RabbitMq/RabbitMqConfigurationExtensions.cs
@@ -22,7 +22,7 @@ namespace Rebus.RabbitMq
                 .Register(c =>
                 {
                     var rebusLoggerFactory = c.Get<IRebusLoggerFactory>();
-                    return new RabbitMqTransport(connectionString, null, rebusLoggerFactory, false, true, false);
+                    return new RabbitMqTransport(connectionString, null, rebusLoggerFactory);
                 });
 
             configurer
--- a/Rebus.RabbitMq/RabbitMqConfigurationExtensions.cs
+++ b/Rebus.RabbitMq/RabbitMqConfigurationExtensions.cs
@@ -22,7 +22,7 @@ namespace Rebus.RabbitMq
                 .Register(c =>
                 {
                     var rebusLoggerFactory = c.Get<IRebusLoggerFactory>();
-                    return new RabbitMqTransport(connectionString, null, rebusLoggerFactory);
+                    return new RabbitMqTransport(connectionString, null, rebusLoggerFactory, false, true, false);
                 });
 
             configurer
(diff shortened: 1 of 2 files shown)


mookid8000:
I don't think this is enough to support autodelete queues.... if the connection is temporarily lost and then re-established, the queue will be gone and so will any subscriptions that have been made.

In fact, it was such a hassle to make autodelete queues work in Rebus 1 that I intentionally left it out in Rebus 2.

What are your thoughts about this?

mayermd:
I can understand that being a problem and I think you would need to carefully consider it before using an autodelete queue, however, we have some very specific use cases that this could be helpful for and it could result in a lot of queues being spun up, I hate to leave them out there when they aren't being used.  The messages that would be communicated would also not be mission critical, so it wouldn't be the end of the world if they didn't go through.  Would you think there needed to be some sort of functionality to rebuild the queue and subscription if it no longer existed?  Do you have another suggestion on a way to enable temporary communication, maybe even bypassing the rabbitmq?  Maybe I'm making this more complicated than it has to be.

mookid8000:
Yes I do have some suggestions :) you said some things that I think we should consider:

> The messages that would be communicated would also not be mission critical, so it wouldn't be the end of the world if they didn't go through. 

> Do you have another suggestion on a way to enable temporary communication, maybe even bypassing the rabbitmq? 

> Maybe I'm making this more complicated than it has to be.

IMO approaching this problem by taking Rebus - in which we explicitly made an effort to make all operations safe & durable - and then "downgrading" it to no longer be durable and safe because it is not required for this scenario - is attacking the problem the wrong way....

When you say "temporary communication", I'm thinking HTTP. If you want to avoid IIS, a self-hosted HTTP endpoint based on OWIN is _definitely the way to go_. The reasons are many:

* it has the level of durability that you require
* it has a nice model for request/reply
* it is crazy fast
* it does not contribute to making your RabbitMQ server a bottleneck

I usually recommend people to use HTTP for all kinds of request/response scenarios between clients and servers, and then use Rebus to coordinate things with durable messaging between servers.

How do you think that sounds?

mayermd:
Well.  The application I am currently implementing this functionality for is a legacy Winform application that does not currently have a middle tier.  It talks directly to the database.  But, I am trying to set up a bus that would allow clients to talk to each other when certain events occur.  I think my initial approach was wrong however.  I was thinking when a client started performing some action and needed input from other clients if some event occurred, I would spin up a queue with a subscription to messages from other clients.  I think now that I have talked to you and read some more of your documentation, the answer is to create one queue for a client and subscribe/unsubscribe to different message types as necessary.  Does that sound more in line with the architecture you envisioned?  In that case, it wouldn't be that big of a deal to create durable queues as one would only be created for each client.

mookid8000:
Now that you have explained some more about your challenges, I can definitely understand why you were thinking about using RabbitMQ's autodelete queues for this.... and it might actually be one of the few valid use cases for autodelete queues.

My suggestion at this point is this:

If the number of clients is stable, i.e. they do not come and go at random times, I think it would be ok for a client to have its own named persistent input queue, and then you could use Rebus in the client to connect it to the other clients, just like you were thinking.

If on the other hand clients come and go (especially if they often go and never come back again) having persistent queues subscribed to things in RabbitMQ can quickly end up being a huge burden. Therefore, in this case, I would probably use autodelete queues.

If you do choose to use autodelete queues, you can probably make it work with Rebus if you want - you will just need to extend your current solution with the ability to "remember" which things each client is subscribed to, and then be able to re-establish those subscriptions in the event that the client needs to reconnect.

PR #396, opened 2016-01-26
Feature/ravendb sagas

bjomi:
Bug fix where we fix an issue with the bookkeeping of correlation properties that occurred when having a saga that had multiple steps (which is usually the case :)).


mookid8000:
build server is crunching .... it will be out in 0.99.26 some time later today - thanks! :+1:

bjomi:
Great! Thanks!

mookid8000:
and by "later today" I meant that other today that came after yesterday..... (it's out now - thanks for fixing it :grin:)

PR #411, opened 2016-02-12
fix issue with inner exceptions being swallowed when correlation operation for sagas retrieve exceptions

mortenherman:
Changed parameter order of exception initialization to avoid inner exception to be swallowed due to usage of string format without format {}


mookid8000:
good catch! :+1: 

thanks! :smile: 

it'll be out in 0.99.33 in a few minutes

PR #416, opened 2016-02-22
Fixed BasicAzureServiceBusTransport not creating it's own inputqueue

torangel:
Given the folowing code the bus does not create it's own inputqueue. (only an errorqueue)
This seems to be caused by the transport not implementing `IInitializable`

`public async void DoIt()
        {
            var receiverHandlers = new BuiltinHandlerActivator();
            receiverHandlers.Handle<string>(async str => Console.WriteLine("received {0}", str));

            var senderBus = Configure.With(new BuiltinHandlerActivator())
              .Logging(l => l.ColoredConsole())
              .Transport(t => t.UseAzureServiceBus("Endpoint=sb://", "sender",
                        AzureServiceBusMode.Basic))
             .Start();

            var receiverBus = Configure.With(receiverHandlers)
                .Transport(t => t.UseAzureServiceBus("Endpoint=sb:", "receiver", 
                        AzureServiceBusMode.Basic))
                         .Start();
           await senderBus.Advanced.Routing.Send("receiver", "event");
        
        }`

### DIFF
--- a/Rebus.AzureServiceBus/BasicAzureServiceBusTransport.cs
+++ b/Rebus.AzureServiceBus/BasicAzureServiceBusTransport.cs
@@ -16,7 +16,7 @@ using Rebus.Transport;
 
 namespace Rebus.AzureServiceBus
 {
-    public class BasicAzureServiceBusTransport : ITransport
+    public class BasicAzureServiceBusTransport : ITransport, IInitializable
     {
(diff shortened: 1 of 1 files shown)


mookid8000:
Wow, good catch :+1:

mookid8000:
your fix is available in 0.99.35 on NuGet.org now - thanks :grin:

Issue #415, opened 2016-02-19
AzureServicebus transport option to not create queue / check if queue exists


runes83:
Today if the transports checks if via namespacemanager if the queue exists, this requires the client to have manage rights to the servicebus.

We want to be able to have clients that only have read access to a single queue, but this does not work as the bus checks if the queue exists/and creates it when it starts. Would it be possible to add an option to not checking the existence of the queue when starting?

mookid8000:
hmm, I can see that that can be a problem..... and yes, it would certainly be possible to tack some extra information on Azure Service Bus configuration builder, it already supports setting a few options by building on the returned [AzureServiceBusTransportSettings](https://github.com/rebus-org/Rebus/blob/master/Rebus.AzureServiceBus/Config/AzureServiceBusTransportSettings.cs), e.g. like so:

    Configure.With(...)
        .Transport(t => t.UseAzureServiceBus(connStr, queue)
                             .EnablePartitioning()
                             .AutomaticallyRenewPeekLock()
                             .EnablePrefetching(100))

so something like

    Configure.With(...)
        .Transport(t => t.UseAzureServiceBus(connStr, queue)
                             .DoNotCreateQueues())

ought to do the trick

torangel:
If  the readonly instance of Rebus configured with `t.UseAzureServiceBus` is provided with a readonly SAS for it's own "readonly" queue (and not for the namespace) it will not be able to send messages to other queues in the namespace. Would it be better with the folowing configuration to clearly indicate that this bus-instance will never be able to send messages (except to itself)?:

`Configure.With(...)
    .Transport(t => t.UseAzureServiceBusReadOnly(connStr, queue)
                        )`

torangel:
Could this work?
#417

mookid8000:
hi @torangel ,

I am sorry, but it's not clear to me what this does?

Or maybe it's the naming that is confusing me....

If I understand correctly, SAS is a tailored connection string (similar to a JWT/SWT/etc), i.e. basically an encrypted set of claims that can be used to give granular access rights to something.

In this case - and this is where it becomes unclear to me - you're calling `UseAzureServiceBusReadOnly` with a SAS connection string that grants the endpoint the right to RECEIVE and SEND on its own queue only, right?

Or how do you use the SAS connection strings?

torangel:
Sorry for being unprecise with the naming here. My apologies.

[...]

PR #433, opened 2016-03-25
Now possible to configure whether ASB-queues should be automatically …

torangel:
This makes it possible to use rebus as "receiveonly" when using a shared
access policy with "listen" permissions

Ref:
#415 

(there will still be an issue with the errorqueue if the bus is configured to use a shared access policy defined only for the input queue)


mookid8000:
looks great :+1:  - will be out in 0.99.44 in a few seconds :grin:

PR #450, opened 2016-05-18
Clean-up + use the c#6 function "nameof()". So if you change the variable you will get compilation errors

SvenVandenbrande:
---


mookid8000:
thanks again 👍

Issue #446, opened 2016-05-17
Sql Server transport and storage lacks SqlConnection costructor


zabulus:
We are using custom configuration through registry with encryption involved for our database. But I'd like to use our SqlServer as SqlServer centralized Subscription Storage. 

I suggest to add following method to class SqlServerPersistenceConfigurationExtensions and maybe to similar ones:
```public static void StoreInSqlServer(this StandardConfigurer<ISagaSnapshotStorage> configurer, Func<SqlConnection> connectionFactory, string tableName, bool automaticallyCreateTables = true)```

zabulus:
I see EnsureMarsIsEnabled function in the DbConnectionProvider. 
1. Is MultipleActivityResultSet is mandatory for Sql storages?
2. Is it okay if I check (and patch?) for MARS in GetConnection function instead DbConnectionProvider.Ctor?

mookid8000:
1. Well... because of the lazy nature of many of the things that the connection is going to be used for, it is required, yes. Or it used to - come to think of it, it might not be required anymore. I will have to do some testing to figure this out.
2. Yeah sure - if it turns out that MARS is still required, I can see no problem in moving it... although the best way would probably be for the `DbConnectionProvider` to build the MARS-enabled connection string in the ctor and keep it, so it doesn't have to do it every time it is needed.

mookid8000:
fixed by #454

mookid8000:
Will be out in 0.99.66

PR #454, opened 2016-05-20
Fix Sql Server transport and storage lacks SqlConnection costructor #446

zabulus:
---

### DIFF
--- a/Rebus/Persistence/SqlServer/DbConnectionFactoryProvider.cs
+++ b/Rebus/Persistence/SqlServer/DbConnectionFactoryProvider.cs
@@ -1,0 +1,108 @@
+﻿using System;
+using System.Configuration;
+using System.Data.SqlClient;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Transactions;
+using Rebus.Logging;
+using IsolationLevel = System.Data.IsolationLevel;
+
+#pragma warning disable 1998
+
+namespace Rebus.Persistence.SqlServer
+{
+    /// <summary>
+    /// Implementation of <see cref="IDbConnectionProvider"/> that ensures that MARS (multiple active result sets) is enabled on the
+    /// given connection string (possibly by enabling it by itself)
+    /// </summary>
+    public class DbConnectionFactoryProvider : IDbConnectionProvider
+    {
+        readonly Func<SqlConnection> _connectionFactory;
+        readonly ILog _log;
+        bool _loggedAboutMars = false;
+
+        /// <summary>
+        /// Uses provided SqlConnection factory as constructor for SqlConnection used. Will use <see cref="System.Data.IsolationLevel.ReadCommitted"/> by default on transactions,
+        /// unless another isolation level is set with the <see cref="IsolationLevel"/> property
+        /// </summary>
+        public DbConnectionFactoryProvider(Func<SqlConnection> connectionFactory, IRebusLoggerFactory rebusLoggerFactory)
+        {
+            if (rebusLoggerFactory == null) throw new ArgumentNullException(nameof(rebusLoggerFactory));
+
+            _log = rebusLoggerFactory.GetCurrentClassLogger();
+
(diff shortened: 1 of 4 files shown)


mookid8000:
I don't mean to be troublesome, but did you consider adding a new implementation of `IDbConnectionProvider`? 

I originally designed it with an interface in place in order to be able to support this exact scenario, so it would be nice if it could make use of that somehow.

zabulus:
Sure no problem.

I have off topic question: why do IDbConnectionProvider and DbConnectionProvider have public visibility? There is no possibility to use user-defined descendants of this classes in rebus.
Do you mind if I change their visibility to internal?

mookid8000:
Are all the SqlServer* persistence implementations internal too?

Sendt fra min iPhone

zabulus:
Nope, all existing implementations have public visibility.

mookid8000:
Ah, good :) the interface should stay public then.... in fact I think that your requirement of supplying the connection from the outside can be implemented without changing anything in the existing implementations, by adding a couple of configuration extensions 

Sendt fra min iPhone

zabulus:
> in fact I think that your requirement of supplying the connection from the outside can be implemented without changing anything in the existing implementations, by adding a couple of configuration extensions

That's exactly what I've added in the PR.

mookid8000:
hmm, I'm looking over your code now... as far as I can tell, the `DbConnectionFactoryProvider` will NOT let you use an already open `SqlConnection` provided from the outside because it will always try to open the connection itself.... I think this pretty much defeats the purpose of being able to supply the connection from the outside in the first place, or is there something I am not getting here?

zabulus:
Indeed.
I can add cheking state of SqlConnection returned from factory method.

mookid8000:
Checking whether the `SqlConnection` was open, and then opening it if necessary, would definitely be one way to do it.

However I'm in doubt as to how useful the `DbConnectionFactoryProvider` would be in its current form. In my opinion it should accept the connection returned by the `Func<SqlConnection>` to be in a usable shape and not do anything to it.

In fact - the more I look at it, the more convinced I am that the `Func` should be `Func<IDbConnection>` because then whoever supplies the `IDbConnection` has more control over what should happen.

For example, `DbConnectionWrapper` can be returned as the `IDbConnection`, which would allow for running both with and without transactions, as well as allowing for managing the connection+transaction externally.

What do you think?

zabulus:
As library user I don't want to be bound to ```IDbConnection```, I have my Sql connection and don't want any additional work to do (inherit from IDbConnection, wrap SqlConnection in it). So I'd like not to change public API side of the PR.

As for your concern for ```DbConnectionFactoryProvider```, I think I can extract all support logic to custom ```IDbConnection``` inheritor, and open connection in it.

mookid8000:
>> As library user I don't want to be bound to IDbConnection, 

You won't. Surely you can base your code on `SqlConnection`, you just pass it to Rebus in something that implements `IDbConnection`.

>> I have my Sql connection and don't want any additional work to do (inherit from IDbConnection, wrap SqlConnection in it). So I'd like not to change public API side of the PR.

Just pass the connection to Rebus using the built-in `DbConnectionWrapper`.

_Additional note:_ if the connection has a started transaction on it, only passing `SqlConnection` won't work because every `connection.CreateCommand()` needs to be followed by a

    command.Transaction = currentSqlTransaction;

and the connection does not provide it anywhere (*). This is what `IDbConnection` should do when you use it to create a command, and it is exactly what `DbConnectionWrapper` does when you initialize it with both a connection and a transaction.


---
(*) This is so stupid that I don't have words for it - why doesn't the connection just set the `Transaction` property to the current `SqlTransaction` automatically when you call `CreateCommand()`?? it doesn't... that's just stupid

mookid8000:
Finally got around to merging this one - sorry about the delay 😁 

Thanks for contributing! 👍

mookid8000:
Will be out in 0.99.66

Issue #393, opened 2016-01-22
Is there any wayt to set priority for message in MSMQ?


alkampfergit:
Probably I've not found in the documentation, but I did not find a way to send a message changing priority in Rebus when I'm using MSMQ.

It would be nice to have an extension that looks in the header for a specific value that can set the priority of that specific message in MSMQ.

alkampfergit:
One of the problem implementing that feature could be that is seems that rebus always send message with a transaction, while [MSMQ does not allow priority to be set if you are using transactions])https://social.msdn.microsoft.com/Forums/windowsdesktop/en-US/92d2ea87-594c-4f9c-8286-35bf21518002/transactional-priority-msmq-sometime?forum=msmq) (one of the reason why you want to move away from MSMQ) ...

mookid8000:
hmmm, well... MSMQ is the only transport that actually has transactions... I've been contemplating moving to the same model as the other transports by storing outgoing messages in-mem until after the user's code has finished successfully...

do you have any thought on this?

alkampfergit:
It should be an opt-in option. 

To avoid breaking existing code, with standard settings everything works as is now, just add another options for MSMQ configuration that allows users to work with Non transactional Queue.

mookid8000:
Yes, you are right. Current MSMQ users will have transactional queues all over the place.

From the link you sent, it looks like you _can_ send transactional messages with priority - you just have to use the same priority for all of the messages.

If the sent messages were collected inmem, the sending of the outgoing messages after running the user's code could consist of grouping the outgoing messages by priority (without reordering them), and then sending them in one transaction per priority.

alkampfergit:
Sadly enough it is not possible :(, as you can read from [this link](http://microsoft.public.msmq.programming.narkive.com/0Qrq4uDm/always-lowest-priority-when-sending-to-transactional-queue). It is a Design decision for MSMQ, if the queue is transactional, always the message will have priority set to Lowest :(.

I think that the only solution is a configuration for MSMQ that allows the user to opt-in for a non transactional queue, then we can add an header for MSMQ to setup priority of the message :).

Gian Maria.

mookid8000:
I think this modification should start its life in a separate package, and then we can merge the implementations later. I would rather not tamper with the current MSMQ transport.

mookid8000:
I'm thinking that this issue qualifies as "up-for-grabs" because it can be made easily by

- [ ] creating a new class library
[...]

PR #465, opened 2016-06-05
#393 feature/non transactional msmq

Martz94:
This is sort of my first time playing with a service bus so just bear with me... :-P

I've tried to follow your check list from [#393](https://github.com/rebus-org/Rebus/issues/393) as closely as possible. There are a few concerns I'm not too sure about though, so I've added comments and `TODO`'s for you to check out. 

Also, 3 tests are failing as they are testing for transactional behaviour (which is gone now when using this new transport)... Didn't know whether to delete the tests or do something else.

Anyway, hope that actually doesn't give you too much work... Happy reviewing!

---


mookid8000:
I haven't had time to look at your PR just yet - sorry about that ... I'll do it, I promise 😄 it'll probably be a few days before I have time though

Martz94:
No worries, take your time.

mookid8000:
Hi @Martz94  -  I am so sorry that I haven't had time to really look at your contribution until now.

As this was a fairly large PR, and due to the fact that the transports are the most central thing in Rebus, I wanted to be pretty thorough when reviewing it.

Looking though your code, I realize that MSMQ unfortunately does not seem to be able to do what is required in order to make this work. Let me explain what this means.... 😒 

You actually kind of hit the nail on the head when you said this:

> 3 tests are failing as they are testing for transactional behaviour (which is gone now when using this new transport)... Didn't know whether to delete the tests or do something else

because _none of the contract tests should be failing_.

The thing is that Rebus makes a promise which it keeps throughout, no matter which transport you choose: It will adhere to a strict [at-least-once delivery guarantee](https://www.google.dk/search?q=at-least-once+delivery).

This means that any errors while handling a received messages should result in the message being put back in the queue... and "any errors" can be anything from a `NullReferenceException` in the handler code to "someone tripped in the power cable and caused everything to blackout".

This should be the case, even with non-transactional MSMQ too - but it seems like it is not possible to achieve, because `messageQueue.Receive(timeout)` does a full receive of the message. There is simply no way to have the message put back in the queue, except, well, actually _manually putting it back in the queue_ (but we would not be around to do that if we suffered a complete, sudden loss of consciousness - with non-transactional MSMQ we would have lost the current message when we woke up).

The way other transports achieve this without the use of "transactions", is by granting a lease on a received message, thus rendering it invisible to other clients. If the receiver then ACKs the message before the lease ends, that means that the message was properly handled. If not, then the message will become visible to other clients again when the lease ends.

The only way to do a transactional receive with MSMQ, it seems, is to create a transactional queue and then supply a `MessageQueueTransaction` to it. But then we're back at the beginning, because we would have to use transactional queues throughout, and that is exactly how the old MSMQ transport works.

I am sorry that it seems like you have done so much work in vain. I really hope you will consider maybe looking at some of the other issues and send another PR 😄

Martz94:
Ok, I see.. Thanks for the detailed explanation!

Issue #478, opened 2016-06-23
There should be a way to propagate the current principal through the message header.


jeffreyabecker:
Really this is a couple of questions:
1. What is the appropriate pipeline stage for outgoing steps? Currently I have this after AssingTypeHeaderStep.
2. Are pipeline steps like owin middlewares in that once await next() has completed we're on our way out? If so is before ActivateHandlersStep the correct place?

mookid8000:
1. Since the job carried out by the step is to add a header to the message, I feel it would be best to put it near the other header steps - so yes 👍 
2.Yes, they're exactly like OWIN middlewares. But I would put it hear the other header-steps.

mookid8000:
An extension method to provide this functionality could be a neat addition to an upcoming convenience package... something like

	Configure.With(...)
		.(...)
		.Options(o => {
			o.AutomaticallyPropagateCurrentPrincipal();
		})
		.Start();

which would automatically

1. Pick up `Thread.CurrentPrincipal.Identity.Name` and put it on outgoing messages, and
1. Set `Thread.CurrentPrincipal` based on the username in a handled message

would actually be pretty awesome.

jeffreyabecker:
Its actually a little more involved than just picking up the name, at least for my usages.  I end up needing to serialize the whole ClaimsPrincipal.Current instance and use that. 

Code is here: https://github.com/jeffreyabecker/Rebus/tree/convenience/Rebus.Recipes/Identity

mookid8000:
ah yes, using `ClaimsPrincipal` is much better

the good looks good, except for one thing: [this part](https://github.com/jeffreyabecker/Rebus/blob/convenience/Rebus.Recipes/Identity/RestorePrincipalFromIncomingMessage.cs#L30-L34) should

1. only do stuff if the `CapturePrincipalInOutgoingMessage.PrincipalCaptureKey` key is present among the headers, and
1. `try { await next(); } finally { revert the principal }` because exceptions will bubble right out of `next()`

I'm looking forward to the recipes PR 😄

mookid8000:
PS: you're on fire! 🔥 who are you? have you been using Rebus for a long time?

jeffreyabecker:
I've been using service buses in the dumbest configurations for years. I used the 'old' version pretty extensively for one app.

jeffreyabecker:
Thanks!

Issue #474, opened 2016-06-23
There should be a way to route every message to a single queue.


jeffreyabecker:
I often rebus in a website to send processes off into the background. When I do this I configure single receiving queue. There should be a simple way to set this up.

mookid8000:
More simple than:

    var bus = Configure.With(...)
        .(...)
        .Routing(r => r.TypeBased().MapFallback("gets_all_the_messages"))
        .Start();

    // ....
    
    await bus.Send("here you go!");

or is that not what you meant?

jeffreyabecker:
Basically just some syntax sugar around that. If you're open to a PR.

mookid8000:
I'm always open to PRs 😄 

Rebus does not have a place yet for things that are basically shortcuts to other stuff that already exists. There's a solution folder called "convenience" though, which is for things that are slightly out-of-band.

Could we start out by adding a new project to the "convenience" folder? Don't know what it should be called.... I just thought about something like `Rebus.Recipes` or something like that - what do you think?

jeffreyabecker:
I like that.

Issue #473, opened 2016-06-23
Fully featured persistence mechanisms should be easier to configure


jeffreyabecker:
For persistence mechanisms that support a full set of interfaces there should be a one-stop configuration mechanism. E.g.

`Configure.With(myioc).UseSqlServer("MyConnection");`

mookid8000:
That's a pretty good idea, although - based on my own experience - I would never get to use one of those (except maybe for demo purposes, which - come to think of it - could be reason enough)

Over time I have converged towards a style where new projects are started by identifying 2-3 endpoint roles, like e.g. "client", "server", and "router", and then provide a custom extension for each... could be something like:

    Configure.With(new CastleWindsorContainerAdapter(container))
        .AsServer("queueName")
        .Start();

which then does all the necessary configuration. This way, the configurations are consistent, while still providing a way to customize each endpoint since `AsServer` just attaches to `RebusConfigurer`

I'll have to think some more about this 😄

jeffreyabecker:
Its actually my most common configuration. Mostly I use rebus as a way to background tasks for my website. Maybe this belongs in that convenience project?

mookid8000:
Yes I think so - but "convenience" is just a solution folder iirc, but you could definitely create the Rebus.Convenience project now

Sendt fra min iPhone

jeffreyabecker:
One open question on this: For Use???AsOneWayClient methods, should I configure saga persistence? I know the old version had issues with that configuration but i havent dug far enough in to tell with the new stuff.

mookid8000:
Since one-way clients don't have an input queue or any worker threads, and thus are unable to receive messages, saga persistence is not relevant.

PR #479, opened 2016-06-24
Rebus.Recipes, #478,  #474,  #473

jeffreyabecker:
This covers several small issues I opened.  
---


Rebus.Recipes/Configuration/CaptureIdentityConfigurationExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+            configurer.AutomaticallyPropagateCurrentClaimsPrincipal(new DefaultClaimsPrinicpalSerializer());
+        }
+        /// <summary>
+        /// Propagates the ClaimsPrincipal through the message bus so that its there during message evaluation
+        /// </summary>
+        /// <param name="configurer"></param>
+        /// <param name="serializer"></param>
+        /// <returns></returns>
+        public static void AutomaticallyPropagateCurrentClaimsPrincipal(this OptionsConfigurer configurer,
+            IClaimsPrinicpalSerializer serializer)
```
mookid8000: wondering why you are providing the serializer instance from the outside? in most cases (if not all) I would imagine that the serialization format of the claims would be pretty unimportant

it's not that this is a problem, but as a general principle I would like to always offer as few overloads as possible

jeffreyabecker: Because WindowsIdentity instances need to use System.Runtime.Serialization. However, I can totally see an argument for making people roll their own in that case.

jeffreyabecker:
I've addressed your concerns regarding the overloads.

mookid8000:
excellent work - thanks! 😄

PR #481, opened 2016-06-27
Fix #480

### LINKED ISSUE #480: AzureStorageQueuesTransport -- Errors with IBus.Defer
When using the AzureStorageQueuesTranasport calling IBus.Defer with an overly short delay can result in a negative timespan being passed to the azure api method. This blows up.

jeffreyabecker:
This addresses issue #480. For short timeouts, the defer window calculation can result in a negative timespan which in-turn blows up the send. This adds bounds checking and returns null if the timespan is less than or equal to zero.
---


mookid8000:
beautiful! 👍 😁

zabulus:
```public static TimeSpan? GetQueueVisibilityDelayOrNull```
public is added here to make visible method for tests. @mookid8000, do you consider to use InternalsVisibleTo attribute?

mookid8000:
ah, good catch - I didn't notice that when I merged the PR

I've made it internal now 👍

PR #524, opened 2016-08-25
Added ASP.NET Core ServiceProvider ContainerAdapter implementation and tests

enosrecanati:
---


mookid8000:
skaermbillede 2016-08-25 kl 10 17 10 😌

mookid8000:
@enosrecanati one of the tests for the new container adapter is failing:

image

The issue is that the container adapter cannot really be disposed in any (at least so it seems) – and Rebus really wants to be disposed when the application shuts down, which is usually ensured by having whichever IoC container is holding the bus singleton instance dispose it.

What is the preferred way to handle disposal with ASP.NET Core ServiceProvider?

enosrecanati:
Good question.

In fact the IServiceCollection implementation is supplied by the runtime, and the registered services disposal is managed by the ServiceProvider itself https://docs.asp.net/en/latest/fundamentals/dependency-injection.html#service-lifetimes-and-registration-options

The bus is registered as Singleton in the Rebus ContainerAdapter, in the docs, is reported that you https://docs.asp.net/projects/api/en/latest/autoapi/Microsoft/AspNetCore/Hosting/IApplicationLifetime/ and in the Startup.cs (where the services as IBus are registered), you can subscribe to the ApplicationStopped and ApplicationStopping events to perform a "graceful shutdown".

mookid8000:
Ah... so what do you think would be the nicest way to be notified on ApplicationStopping?

As in: I am thinking that either

a) `NetCoreServiceCollectionContainerAdapter` should require a second argument in its constructor – a reference to whatever it requires to hook into the ApplicationStopping event

or

b) `NetCoreServiceCollectionContainerAdapter`'s constructor should be changed into requiring as argument one single thing that allows for it to both register itself as `IBus`, look up handler instances, and hook into ApplicationStopping.

Do you think that sounds reasonable?

enosrecanati:
Yes I agree, both solutions seems working to me...I'll do some tests to find out what could be the best solution.

enosrecanati:
I've made some tests, and is not mandatory to pass and instance of IApplicationLifetime to the Container adapter, because it's already registered by there runtime (more precisely from the WebHostBuilder https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs) during the startup.

So the `NetCoreServiceCollectionContainerAdapter` can resolve the interface in his Ctor and hook into `ApplicationStopping` event to dispose the Bus.

I've tried this way and all is working as expected. If it is reasonable I can submit the fix.

PR #529, opened 2016-08-30
Improve `PerformExpiredMessagesCleanupCycle` inside of `SQLServerTran…

xenoputtss:
---

…sport ` so that it doesn't select of a non-index column.


mookid8000:
Very nice 👍

Issue #435, opened 2016-03-26
Ensure non-awaited async stuff runs to end


mookid8000:
When inside a message handler, if you

```
bus.Send(yourMessage);
```

your message might not be properly sent because the continuation ends up being run AFTER the queue transaction has ended (note the lack of an `await`)

MassTransit magically "captures" the started tasks somehow, ensuring that they are run to end before ending the queue transaction. It would be neat if Rebus could do the same.

colin-higgins:
Seeing as you have control over the sending methods, can't you just add the `Task` for the Enqueue action to a shared context? You can await those Tasks after the handler method finishes.

mookid8000:
Dude!! 👏👌👍 that's actually pretty brilliant! I can't believe I did not consider doing that... whenever I thought about this issue, I kept focusing on the synchronization context, but you are absolutely right. 

This will of course only work with Rebus' own tasks... but that's also the most important part in this context. 

Sendt fra min iPhone

colin-higgins:
That's where my mind first went as well... Like.... 
"how the hell could they be capturing tasks internal to a method... Maybe some IL weaving magic??? Wait a second... that's not even necessary."

I guess the nice part about doing this is if users don't "want" to use async, which I understand... but then it's just programming by coincidence.
In other words, the user might expect that the messages will be queued before the rest of their code executes, but that's just not true.

But it does remove blame from Rebus for "losing" messages.

Are you going to go ahead with it?

mookid8000:
I am contemplating making a little spike, yes 😁

mookid8000:
are you asking because you would be interesting in doing it?

colin-higgins:
I'll spike this, yeah.

JornWildt:
I would rather not have this happening magically somewhere/sometime. I would rather see a `WaitAll()` somewhere for the application to call when it is ready.

mookid8000:
I'm closing this one for now, because I don't see how it can be done. Feel free to revive this issue if you have anything to add 😄

PR #549, opened 2016-10-27
[WIP] ConcurrentQueue to await non-awaited rebus tasks post handler

colin-higgins:
---

Rebus is [MIT-licensed](https://opensource.org/licenses/MIT). The code submitted in this pull request needs to carry the MIT license too. By leaving this text in, **I hereby acknowledge that the code submitted in the pull request has the MIT license and can be merged with the Rebus codebase**.

@mookid8000 
Thoughts on this kind of approach? 
For https://github.com/rebus-org/Rebus/issues/435

I need to test and whatnot but I figured I'd get your thoughts first. It was the least obtrusive way I could think of.  (Again, haven't tested yet)
Needs some cleanup yet too.

### DIFF
--- a/Rebus/Activation/BuiltinHandlerActivator.cs
+++ b/Rebus/Activation/BuiltinHandlerActivator.cs
@@ -126,16 +126,35 @@ namespace Rebus.Activation
         {
             readonly Func<IBus, TMessage, Task> _handlerFunction;
             readonly Func<IBus> _getBus;
+            IBus _shortLivedBus = null;
 
             public Handler(Func<IBus, TMessage, Task> handlerFunction, Func<IBus> getBus)
             {
                 _handlerFunction = handlerFunction;
-                _getBus = getBus; // store this function here because of Hen&Egg-Problem between handler activator and bus
+                // store this function here because of Hen&Egg-Problem between handler activator and bus
+                _getBus = () =>
+                {
+                    _shortLivedBus = getBus();
+                    return _shortLivedBus;
+                };
             }
 
             public async Task Handle(TMessage message)
             {
                 await _handlerFunction(_getBus(), message);
+                await AllUserTriggeredTasks();
+            }
+
+            private async Task AllUserTriggeredTasks()
+            {
+                if (_shortLivedBus != null)
+                {
+                    var rebusImplementation = _shortLivedBus as RebusBus;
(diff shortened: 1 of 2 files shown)


mookid8000:
Definitely great to see an idea on how it could be done, but as far as I can tell by the work you have done, it would only work with `BuiltinHandlerActivator`... or am I missing something?

You seem to have done one thing though, which I think is pretty brilliant: The `EnqueueTo` extension method, which I think could maybe be changed to something like `EnsureIsAwaited` and look somewhat like this:

	public static EnsureIsAwaited(this Task task) 
	{
		var context = AmbientTransactionContext.Current;
		if (context == null) return; //< we are not in a Rebus handler

		var taskQueue = context.Items.GetOrAdd("tasksToFinish", () => {
			var tasks = new ConcurrentQueue<Task>();
			context.Committed += async () => await Task.WhenAll(tasks);
		});

		taskQueue.Enqueue(task);
	}

(which is written purely from memory – it probably can't compile)

What do you think?

mookid8000:
BTW it looks like you're working on an old fork of Rebus – in case you make another attempt, you should fetch upstream before you start working 😉

colin-higgins:
Yeah it is quite old. I had pulled it down and started scratching around a few months ago. :)

I didn't want to find every point of entry until I vetted the idea.

I very much like your method, I hadn't thought of using the transaction context.
I'll get latest and fork again tonight my time.

mookid8000:
🤘

colin-higgins:
PS, is there ever a situation where there would be no transaction context within a handler?

mookid8000:
> PS, is there ever a situation where there would be no transaction context within a handler (or a future situation)?

No.

> Is it conceivable?

No! 😁   (1)

> And do we really want to wait until it is committed? That sounds too late to me.

I remembered wrong. The API looks like this:

    context.Commit(async () => {
        // async stuff in here
    });

and it is meant for registering actions that must complete BEFORE acknowledging that the message has been properly handled.

IOW it is the right time 😄 

---

1) All good rules have an exception. Since you are asking whether you can ever encounter a situation where there is no transaction context **within a handler**, there is actually one that I can think of: When calling `await Handle(someMessage);` from within a unit test. But the `RebusBus` does not have to account for that in any way.

mookid8000:
Definitely an interesting experiment, but I'm closing this PR for now.

PR #570, opened 2016-12-26
Feature/mvandevy

mvandevy:
Created a .net standard 1.6 compliant version of the Rebus library. Important remarks: could only make it compatible with .NETFramework 4.6 as the minimum & had to convert the projects to .xproj/.json file combinations. So the .csproj are gone. This thus only works with VS 2015 with Update 3. All unit tests run on Windows and MacOS. Haven't tested them on linux yet though.

Had to introduce a compiler variable .net46 for the code that is only .NETFramework 4.6 compliant. Another downside is the $version$ variable that I haven't found a solution for. The version number is set specifically in the project.json files and the Assembly.cs files for the moment.

---


seankearon:
No big shakes, but I'm wondering why I'm getting CLA emails for this?

mvandevy:
No idea either, but I wasn't able to create the PR without complying to the CLA. Maybe related to that?

mookid8000:
Super 😄 I have a couple of questions though:

> could only make it compatible with .NETFramework 4.6

why? isn't it simply a matter of targeting netstandard1.1 (as per the table listed in [this blog post](https://blogs.msdn.microsoft.com/dotnet/2016/09/26/introducing-net-standard/))

Even though 4.5 is no longer officially supported by Microsoft (4.5.2 is the lowest supported .NET version to my knowledge), Rebus needs to be available on 4.5 still. I am also fairly certain that that should be possible.

> had to convert the projects to .xproj/.json file combinations. So the .csproj are gone

I haven't been following along properly, so my knowledge might be slightly off – but I think .xproj + project.json is "the old format", as in: If you update to VS 2017 (which in in RC now) I think you can have the new .csproj format that works with .NET Core and all that stuff.

mvandevy:
> could only make it compatible with .NETFramework 4.6
> why? isn't it simply a matter of targeting netstandard1.1 (as per the table listed in this blog post)

I started out targeting 4.5 and netstandard1.1, but netstandard1.1 only specifies a smaller subset of the APIs available in 4.5. The best explanation I can give is by pointing you to the 'developer explanation' of David Fowler (https://gist.github.com/davidfowl/8939f305567e1755412d6dc0b8baf1b7). As it explains, netstandard1.1 lacks the Threading.Timer, FileSystem, Threadpool, ... APIs making it hard to impossible to target that version of netstandard. The minimum version I could target was netstandard1.3 which is only compatible with .NETFramework 4.6

> Even though 4.5 is no longer officially supported by Microsoft (4.5.2 is the lowest supported .NET version to my knowledge), Rebus needs to be available on 4.5 still. I am also fairly certain that that should be possible.

I spent a quite a lot of time looking for a solution to keep it compatible with 4.5, but could not find a suitable solution due to the missing APIs in netstandard1.1. If anyone knows a solution, or can point me in the direction of solution, then please do so. Already spent quite some time on this, but couldn't get it to work. 😕 You're probably targeting 4.5 because you don't want to impact the projects already running Rebus on 4.5 which can't be upgraded easily, or is there another reason?

> had to convert the projects to .xproj/.json file combinations. So the .csproj are gone
I haven't been following along properly, so my knowledge might be slightly off – but I think .xproj + project.json is "the old format", as in: If you update to VS 2017 (which in in RC now) I think you can have the new .csproj format that works with .NET Core and all that stuff.

Indeed, the .xproj + project.json is the old format, but wasn't sure if it was suitable to have people upgrade to VS 2017, even if it is a RC. If that is what is needed to have a smoother transition, then I'm happy to upgrade 😄

mookid8000:
> You're probably targeting 4.5 because you don't want to impact the projects already running Rebus on 4.5 which can't be upgraded easily, or is there another reason?

Well – it's because .NET 4.5 is the lowest version of .NET that has all of the necessary features (mainly Tasks and support for `async`/`await`) thus allowing Rebus to come out in one single DLL that can be consumed on all subsequent .NET platforms.

Isn't it possible to simply put `net45` as the target?

> If that is what is needed to have a smoother transition, then I'm happy to upgrade

It would be lovely 😄 I am running solely on VS 2017 now.

mvandevy:
Ok, some progress on this. Been a bit preoccupied lately so wasn't able to continue on this. I've move the solution over to VS 2017 RC, but I won't be able to get it working for .NET 4.5. Simply putting `net45` doesn't do the trick because I had to refactor some code. The code I'm referring to is i.e. `AmbientTransactionContext` which makes use of the `CallContext` type. This is something that is not available in .net standard (and probably won't become available due to multiple platforms being targetted: win, linux, mac). As an alternative I had to use `AsyncLocal`, but this is only available from `net46` onwards. So the minimum compatibility will be `net46` unfortunately. I'll try to update the code asap.

mookid8000:
> The code I'm referring to is i.e. AmbientTransactionContext which makes use of the CallContext type. This is something that is not available in .net standard (and probably won't become available due to multiple platforms being targetted: win, linux, mac). As an alternative I had to use AsyncLocal, but this is only available from net46 onwards

It should be possible to fix this using conditional compilation directives – if MSBUILD does not do it automatically, it should be possible to define a symbol with the target framework moniker for each build, so that e.g. `NET45` is defined when compiling for .NET 4.5.

This way, source code can do this:

    #if NET45
    CallContext.LogicalSetData(...);
    #else
    AsyncLocal.(I don't even know what this is)
    #endif

which is OK when used sparingly.

mvandevy:
Indeed, but was trying to convert it without conditional compilation directive 😁 Seems I'll have no other option.

PR #585, opened 2017-02-28
Feature/mvandevy

mvandevy:
I was previously unable to get it building for the multiple targets. Probably related to my system setup and/or version of VS 2017. In short, I always bumped into issues. Upgrading to newer RC versions helped me tackle some of these issues, but other ones arose or were reintroduced which caused me to basically give up. Now that we are near to the actual release of VS2017 I thought to give it another shot. Behold, I seem to have succeeded in getting Rebus to target both netstandard 1.6 and .net 4.5. 😄 

So in short, what actually changed: 

1. csproj files changed to use toolsversion 15 with the Microsoft.NET.Sdk. This still has some impact though for your workflow. Let me walk over the changes:
	* the csproj file now includes the packaging information previously found in nuspec files. The biggest change is the inability to use the $version$ variable here. Haven't really investigated this further, but the short term fix could be to manually keep them up to date. It also seems that the UI of VS2017 is not picking up the version correctly, so you'll manually have to edit the csproj file if you want to update the version.
	* The nuspec file also contained a ``<files>`` element, and I haven't been able to look into this yet.
	* where you previously were able to execute a 'patch_assemblyinfo.cmd' command by using a PreBuildEvent, this had to be changed to a ``<Target Name="PreBuild" BeforeTargets="PreBuildEvent">``. When I kept it in the PreBuildEvent, then the variables, like $(ProjectDir) were not getting interpreted and replaced by their actual data. You won't be able to edit the command from the UI in 2017. You'll need to manually modify this command in the csproj file when changes are needed.
	* Had to add the ``<GenerateAssemblyInfo>false</GenerateAssemblyInfo``>`` tag to the csproj 
2. Other code changes:
	* Replaced the usage of System.Timers.Timer with System.Timers.Timer
	* Changed the RijndealEncryptor to an AesEncryptor in .net standard
	* Modified the ApplicationException to become RebusApplicationException instances
	* AccessViolationException have become  UnauthorizedAccessException
	* Inherited from Exception instead of ApplicationException in .net standard. ApplicationException is no longer available in .net standard. Inheriting from ApplicationException is even seen as a bad practice.
	* in FileSystemExclusiveLock the filestream is not locked or unlocked as this cannot be done in .net standard. For .net 4.5 the code remains as it was.
	* string comparison uses the OrdinalIgnoreCase in .net standard as InvariantCultureIgnoreCase is not available.
	* JsonSerializer: _encoding.HeaderName has to be replaced with _encoding.WebName which seems to be closest.
	* AmbientTransactionContext: used AsyncLocal instead of CallContext in .net standard.
	* ThreadPoolWorker: used OperationCanceledException instead of  ThreadAbortException in .net standard
---


mvandevy:
It seems AppVeyor can't really deal with the new csproj format yet. Then again, I don't really know AppVeyor, so it could be sth else as well 😃

mvandevy:
Just wondering if there is anything left I can assist with in order to get Rebus compatible with both .net 4.5 and .net standard 1.6?

mookid8000:
I don't think so – you've done pretty much already 😄 

All this is waiting for, is me having the time to properly understand the awesome work you have done – I need to be pretty sure that the .NET Core things work as a pure addition, so that all of the current .NET 4.5 stuff is preserved.

For a change this big, I think I need at least a day or two to properly work it through. I'm sorry, but I can't really make any promises at the moment.

mookid8000:
WTF?! sorry, hit the wrong button!

mookid8000:
argh!

mookid8000:
sorry, I am such an idiot – I could not remember how far I got the last time I looked at this, so I wanted to start over – and now I deleted the PR!

mookid8000:
I will pull your branch manually then.... I am going to look at this today!

PR #615, opened 2017-05-25
PR for #612

Rzpeg:
---

### DIFF
--- a/Rebus.Tests/Extensions/TestTypeExtensions.cs
+++ b/Rebus.Tests/Extensions/TestTypeExtensions.cs
@@ -114,7 +114,7 @@ namespace Rebus.Tests.Extensions
         {
             // arrange
             var expectedType = typeof(ComplexGenericMessage<SimpleMessage, int>);
-            var expectedTypeString = "Rebus.Tests.Extensions.ComplexGenericMessage`2[[Rebus.Tests.Extensions.SimpleMessage, Rebus.Tests],[System.Int32, mscorlib]], Rebus.Tests";
+            var expectedTypeString = "Rebus.Tests.Extensions.ComplexGenericMessage`2[[Rebus.Tests.Extensions.SimpleMessage, Rebus.Tests], [System.Int32, mscorlib]], Rebus.Tests";
 
             // act
             var actualTypeStringStatic = typeof(ComplexGenericMessage<SimpleMessage, int>).GetSimpleAssemblyQualifiedName();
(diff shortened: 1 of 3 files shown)


Rzpeg:
Hey!
Any ETA for v.3 nuget update ?

mookid8000:
> Any ETA for v.3 nuget update ?

Hopefully within a few minutes 😄 

> P.S. We will be moving to .net standard 2.0 when its out, it's great that rebus is being ported to .net standard!

Awesome! Rebus 4 (which targets .NET 4.5 and .NET Standard 1.6) will be out soon – I am currently testing it myself, but it would of course be helpful if more people would try out the 4.0.0-b10 version and report any issues found.

mookid8000:
It's out in 3.1.4 now 😄 excellent work 👍

Rzpeg:
Cool, thanks! 
You will eventually update the v4 branch with the change, won't you ?

Sorry for off-topic, but weren't you working on rebus worker threads autoscale ? Has it been tested already ? How does it relate to SetMaxParallelism ? (I always found it confusing to have both SetMaxParallelism, and SetNumberOfWorkers, because in my use-cases, I wan't a thread to do the maximum possible work. (i.e. SetMaxParallelism - unlimited (does this option exist, by the way ? )))

mookid8000:
> You will eventually update the v4 branch with the change, won't you ?

Yes, thanks for reminding me 😜  – I'll push 4.0.0-b11 in a few minutes!

> Sorry for off-topic, but weren't you working on rebus worker threads autoscale ? 

Yes – it's called [Rebus.AutoScaling](https://github.com/rebus-org/Rebus.AutoScaling)

> Has it been tested already ? 

Not really – I haven't had a scenario yet where I think it made sense.

> How does it relate to SetMaxParallelism ? (I always found it confusing to have both SetMaxParallelism, and SetNumberOfWorkers, because in my use-cases, I want a thread to do the maximum possible work. (i.e. SetMaxParallelism - unlimited (does this option exist, by the way ? )))

"Number of worker threads" is simply the number of real .NET threads that Rebus spins up when starting. If you don't `await` anything (and if the transport itself does not support `Task`-based parallelism, e.g. like MSMQ) this will determine how many things can actually happen in parallel.

"Max parallelism" is the absolute upper bound across all threads of how many messages can at most be processed in parallel. This setting is required, because otherwise _too much_ parallel stuff would simply be going on. E.g. if the transport supports `Task`-based receive, like Azure Service Bus, Amazon SQS, etc. then one single worker thread would immediately initiate 1000s of receive operations until running out of connections or whichever resource is associated with a pending receive operation.

Therefore you will probably always want an upper bound on the number of parallel operations. You can of course experiment freely with these settings and e.g. set the parallelism to some ridiculously high value like `int.MaxValue` to effectively disable the upper bound, but I am pretty sure that that is not what you actually want 😄 

If you really want lots of parallel things going on, and your transport can handle so many messages being processed concurrently, I suggest you set it to something like 50 or 100.

I hope that makes sense 😄 [The wiki page about workers/parallelism](https://github.com/rebus-org/Rebus/wiki/Workers-and-parallelism) has a couple of examples listed on some sensible settings.

PR #623, opened 2017-06-28
Bumping TFM down to "netstandard1.3"

mirandaasm:
Hey bro, sup?

I've managed to bump the target framework version for .NET Standard down to v1.3.

Unfortunately, "STT" (System.Threading.Thread) libs aren't available in .NET Standard v1.0-v1.2, so I'm guessing that 1.3 is the lowest we can go with.

"Environment.MachineName" isn't available either, so I replaced these occurrences with a small "hack" that I found out in a SO question. Not sure if there's anything other than this hack available too.

Other than that, had to bump NUnit version up to 3.7.1, and add some other MS package references.

---

### DIFF
--- a/Rebus.Tests.Contracts/Extensions/Ponder.cs
+++ b/Rebus.Tests.Contracts/Extensions/Ponder.cs
@@ -17,7 +17,7 @@ namespace Rebus.Tests.Contracts.Extensions
 
             foreach(var dot in dots)
             {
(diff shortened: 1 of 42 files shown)


mookid8000:
absolutely awesome work!!

you even succeeded in preserving the pipeline IL compiler, which I had to remove the first time I tried to lower the target 😆

mookid8000:
it's out in 4.0.0-b16

Issue #341, opened 2015-09-04
Saga Concurrent Updates


MrMDavidson:
We're having an issue with Rebus where multiple workers are attempting to work on a saga and they're hitting a `ConcurrencyException`. We're using the SQL transport and saga storage. A reproduction case is included (just update `connectionString` before building) and tested against the current NuGet package (0.98.5). Once the app spins up it'll you'll be presented with 3 options. Hit "1" to add an additional worker. "2" to remove a worker. And "3" to start the saga. To begin with add a single worker and then press 3. It'll queue a `StartSagaCommand` command that initiates a new `CountingSaga`. This simple saga sends 100 `IncrementCounterCommand`s which will be handled by the same saga. When a `IncrementCounterCommand` a counter is incremented in the saga data and if we've received the 100 messages we mark the saga as complete and print a message. You'll see lots of `ConcurrencyExceptions` thrown. If you add another worker (Hit 1) or fire up another copy of this app you'll get even more. To the point where the saga probably won't even finish.

(Side note: If I set `SetMaxParallelism()` to a lower value I have a better chance of success. But this still doesn't help if I have multiple workers. What's the purpose of this? What is a sane value? Here I've set it to 10)

From my understanding what's happening is that Worker Thread A is pulling a message, `LoadSagaDataStep` is executing, seeing it's in a saga, loading the saga data at version N, awaiting the work for that worker, and then updating the saga. In the `SqlServerSagaStorage` `Update(...)` method an `UPDATE ... WHERE ... [Revision] = $CurrentRevision$` is performed and, if no rows are found, it throws a `ConcurrencyException`. As more worker threads are operating the chance of the revision being incremented between loading the saga and updating it increases. 

So, two things...
1. Am I doing something completely wrong here? Admittedly this example doesn't do a great deal of work so multiple messages processing at about the same time seems very likely. But it's hard to get buy in from people if this causes failure.
[...]

PR #632, opened 2017-08-07
Single access sagas

MrMDavidson:
Introduces the concept of Single Access Sagas. These are sagas for which only one handler will execute at once. To enable these;

```C#
RebusConfigurer config = Configure.With(...)
config.EnableSingleAccessSagas()
```

This inserts a `SingleAccessSagaIncominbgStep` into the incoming pipeline before the `LoadSagaDataStep`. It will inspect all of the handlers for the incoming message and see if any of them have `HasSaga == true` and are an instance of `ISingleAccessSaga`. If no such handler satisfies these conditions then the message pipeline continues as normal.

```C#

public class MyCPUIntensiveAndChattySaga : Saga<MySagaType>, ISingleAccessSaga 
{
  // ...
}

```

If one or more handler *does* satisfy these conditions then all of the correlation properties are inspected for each `ISingleAccessSaga` message handler and correlation property. For each correlation property the ``ISagaLockProvider` requests a `ISagaLock` for that property value and the lock is attempted to be attained. If all locks can be attained then message processing continues as normal.

If not all locks were obtained then the message will be deferred for 5 - 10 seconds. **NOTE**: This means that the queue ordering may be disturbed. In practical applications this shouldn't be an issue. Also the time for deferral / strategy for deferral should probably be pushed into a new dependency.

This also comes with a `SemaphoreSagaLockProvider` / `SemaphoreSagaLock` which uses a machine wide semaphore. If your setup has more than one worker machine then you will need to implement and register a your own `ISagaLockProvider` implementation which supports a distributed locking system (eg. Redis Locks, SQL Server App Locks, etc).

For the full context of this see https://github.com/rebus-org/Rebus/issues/341

---

### DIFF
--- a/Rebus/Config/RebusConfigurer.cs
+++ b/Rebus/Config/RebusConfigurer.cs
@@ -227,7 +227,7 @@ namespace Rebus.Config
                 return new SimpleRetryStrategy(simpleRetryStrategySettings, errorTracker, errorHandler);
             });
 
-			PossiblyRegisterDefault(c => new SemaphoreSagaLockProvider());
+			PossiblyRegisterDefault<ISagaLockProvider>(c => new SemaphoreSagaLockProvider());
 
             PossiblyRegisterDefault(c => new SimpleRetryStrategySettings());
 
--- a/Rebus/Config/RebusConfigurer.cs
+++ b/Rebus/Config/RebusConfigurer.cs
@@ -20,6 +20,7 @@ using Rebus.Retry.Simple;
 using Rebus.Routing;
 using Rebus.Routing.TypeBased;
 using Rebus.Sagas;
+using Rebus.Sagas.SingleAccess;
 using Rebus.Serialization;
 using Rebus.Serialization.Json;
 using Rebus.Subscriptions;
@@ -225,6 +226,8 @@ namespace Rebus.Config
                 var errorHandler = c.Get<IErrorHandler>();
                 return new SimpleRetryStrategy(simpleRetryStrategySettings, errorTracker, errorHandler);
             });
+
+			PossiblyRegisterDefault(c => new SemaphoreSagaLockProvider());
 
             PossiblyRegisterDefault(c => new SimpleRetryStrategySettings());
 
--- a/Rebus/Injection/Injectionist.cs
+++ b/Rebus/Injection/Injectionist.cs
@@ -210,7 +210,7 @@ namespace Rebus.Injection
 
                 if (!_resolvers.ContainsKey(serviceType))
                 {
-                    throw new ResolutionException("Could not find resolver for {serviceType}");
+                    throw new ResolutionException($"Could not find resolver for {serviceType}");
                 }
 
                 if (!_decoratorDepth.ContainsKey(serviceType))
--- a/Rebus/Sagas/SingleAccess/ISagaLock.cs
+++ b/Rebus/Sagas/SingleAccess/ISagaLock.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Rebus.Sagas;
+
+namespace Rebus.Sagas.SingleAccess
+{
+	/// <summary>
+	/// Defines an interface for acquiring a lock on a <seealso cref="Saga{TSagaData}"/>. Implementors should release the lock, if acquired, on disposal
+	/// </summary>
+    public interface ISagaLock : IDisposable
+	{
+		/// <summary>
+		/// Attempt to acquire a lock. If the lock was successfully acquired return <c>true</c>. If the lock could not be acquired returns <c>false</c>
+		/// </summary>
+	    Task<bool> TryAcquire();
+    }
+}
--- a/Rebus/Sagas/SingleAccess/ISagaLockProvider.cs
+++ b/Rebus/Sagas/SingleAccess/ISagaLockProvider.cs
@@ -1,0 +1,21 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Rebus.Sagas.SingleAccess
+{
+	/// <summary>
+	/// Interface for providing a lock on a saga
+	/// </summary>
+    public interface ISagaLockProvider
+	{
+		/// <summary>
+		/// Acquire a <seealso cref="ISagaLock"/> that can provide mutual exclusion semantics to a given saga
+		/// </summary>
+		/// <param name="sagaCorrelationId">The correlation identifier of the saga the lock is requested for. Implementors would usually use this to key a lock specific to this saga</param>
+		/// <returns>A <seealso cref="ISagaLock"/> that can be acquired for this saga</returns>
+	    Task<ISagaLock> LockFor(object sagaCorrelationId);
+    }
+}
--- a/Rebus/Sagas/SingleAccess/ISingleAccessSaga.cs
+++ b/Rebus/Sagas/SingleAccess/ISingleAccessSaga.cs
@@ -1,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Rebus.Sagas;
+
+namespace Rebus.Sagas.SingleAccess
+{
+	/// <summary>
+	/// Marker interface used to identify a <seealso cref="Saga{TSagaData}"/> implementation as requiring single handling across all workers
+	/// </summary>
+    public interface ISingleAccessSaga
+    {
+    }
+}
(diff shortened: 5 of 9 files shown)


MrMDavidson:
Just a note on this: One of my original plans was to do this as an add on nuget package - I did start down this path but `SagaHelper` was moved to an internal class and I didn't want to have to duplicate the logic. I think for now, provided you're okay with it, the Single Access Saga would be base functionality of Rebus. With the various add ons (Rebus.SqlServer, etc) potentially providing their own `ISagaLockProvider` / `ISagaLock` implementation

MrMDavidson:
Just noticed there's a weird edge case where I'm still getting concurrency exceptions here - hold off merging this until I can nail down what's happening. I'd still appreciate someone taking a look at this, though, and seeing how you feel about it.

Rebus/Config/RebusConfigurer.cs
```diff
@@ -227,7 +227,7 @@ public IBus Start()
                 return new SimpleRetryStrategy(simpleRetryStrategySettings, errorTracker, errorHandler);
             });
 
-			PossiblyRegisterDefault(c => new SemaphoreSagaLockProvider());
```
mookid8000: 👍

Rebus/Config/RebusConfigurer.cs
```diff
@@ -227,7 +227,7 @@ public IBus Start()
...
+			PossiblyRegisterDefault<ISagaLockProvider>(c => new SemaphoreSagaLockProvider());
```
mookid8000: 👍

Rebus/Injection/Injectionist.cs
```diff
@@ -210,7 +210,7 @@ public TService Get<TService>()
 
                 if (!_resolvers.ContainsKey(serviceType))
                 {
-                    throw new ResolutionException("Could not find resolver for {serviceType}");
+                    throw new ResolutionException($"Could not find resolver for {serviceType}");
```
mookid8000: 👍  good catch!

MrMDavidson:
I've had a bit more of an investigation into the issue I was having and have a pretty reliable test case. Spin up a new C# console project a build of this branch for Rebus and Nuget Rebus.SqlServer. Copy paste https://gist.github.com/MrMDavidson/a7eb762d89855f67974aebc602e8d7e0 as your Program.cs and run it. When you do you'll be greeted with a prompt asking you for your bus mode. Try it with M (`InMemory`) then F (`FileSystem`) then finally S (`SQLServer`). Once you've selected a bus mode hit 1 at least once to spin up a worker thread. Then press 3 to send a normal style Saga. If you have more than 1 worker thread you'll notice it fails for `InMemory`, `FileSystem` and `SQLSeerver` with the saga throwing a `ConcurrencyException`. You can add / remove workers with 1 and 2 and observe how it reacts with different worker counts - more of them and you're more likely to hit `ConcurrencyException`s Now repeat the test bus press 4 to start a saga implemented as an `ISingleAccessSaga`. `InMemory` and `FileSystem` work fine - no matter the worker count it succeeds. But `SQLServer` continues to throw `ConcurrencyException`s. I haven't tried other transports - if you have the ability to it'd be interesting to replicate this test with them. But at this stage I'm guessing that the `SQLServer` transports are doing something strange that mean other works are taking a message even though they... shouldn't be...

MrMDavidson:
I'm at a bit of a loss... I used a different machine today and `SQLServer` as a  transport now works flawlessly. Exact same code running here. Would be good to have someone once over this and see if they're able to reproduce any issues.

mookid8000:
hi @MrMDavidson , I have come to the conclusion that this feature should not be released as part of Rebus core just yet.

I am pretty eager to release Rebus 4 right now, and I think it would be too risky to put this feature in (at this point in time) without a period of proper testing, which would then delay the release quite a bit.

Therefore, I am closing this PR, and then hopefully we can start this feature as its own add-on package and then move it to core for Rebus 5 if it turns out to be good (which I think it will).

I am sorry that we got this far with the PR and all – I'll be happy to help you establish a new repository where we can develop this feature and then hopefully include it as part of core Rebus 5.

MrMDavidson:
That's most unfortunate... I'd hoped to make it part of the core so other people would be able to make use of it.

That said; if it would be possible to partially rollback 673f1d5bb1c5f7404be93d4e026f35496936a468 so that `SagaHelper` was public I could release this as something that sat along side Rebus without needing to be in the core.

mookid8000:
I'll release Rebus 4.0.0-b20 in a few minutes with a public `SagaHelper`

MrMDavidson:
Thanks - I appreciate it.

mookid8000:
it's out now 😄

MrMDavidson:
Cheers - I'll whip up a public package based on this PR tomorrow at work.

MrMDavidson:
Made https://www.nuget.org/packages/Rebus.SingleAccessSagas/ available 🎉

mookid8000:
beautiful!

cldons:
This feature should really find it's way into Rebus 5 core.

Sagas that are open to asynchronously processing but can be configured to be closed to concurrent processing occur much more often than we think. The single access saga's package is a wonderful addition - just like IdempotentSagas - and it's almost a basic requirement for them to have. Why? Because for some sagas we only want to have a single instance processing the messages. 

Actually we can think of having sagas that are long running - as to Udi's definition - as long as the user exists. Like to good old amazon shopping cart example or the example of buying in-app game credits, both examples he provides. Those sagas live for the lifetime the user has an account - we will not even call MarkAsComplete on them. 

In my opinion those kinds of sagas should offer some form of synchronized access to them - allowing for better eventual consistency.

@MrMDavidson great work on the single access saga!
@mookid8000 looking forward to Rebus 5 ;)

mookid8000:
It's a little bit of a secret, but Rebus has actually had single-access sagas in its core since [4.1.0](https://github.com/rebus-org/Rebus/blob/master/CHANGELOG.md#410), in a very simple form that uses a bunch of semaphores to implement exclusive access.

You can enable it like this:
```csharp
Configure.With(...)
    .(...)
    .Sagas(s =>
    {
        s.StoreInSqlServer(...);
        s.EnforceExclusiveAccess();
    })
    .Start();
```

The feature is very much inspired by @MrMDavidson 's work.

It _could_ be extended to support distributed locks too, but for now I've only seen a few use cases, and in-mem locks were good enough for those.

MrMDavidson:
@cldons Glad you find the Single Access Saga package useful, that's awesome to hear 😄 

@mookid8000 In our environment we autoscale the workers processing the queue - we run anywhere from 2 - 20 instances. Some of the stuff they do is very CPU intensive, and long running (minutes, not seconds). So for us we have a need for a distributed locking scheme (in our scenario we're using Redis to provide the locks across workers). I haven't looked too deeply into the core implementation (I did see it was in 4, congrats!) but in the package I provide people can override the implementations of `ISagaLockProvider` to return custom `ISagaLock` implementations. For dependency minimisation I only ship a semaphore based lock - like you - but in house I have a `RedisSagaLockProvider` and `RedisSagaLock` to synchronise the lock across all N worker machines. I imagine it wouldn't be too hard for the core implementation to do this, as well. All just using the built in dependency resolution in Rebus 👍

cldons:
Well thats great news too, I didn't know it was available in the form of Exclusive Access configuration. 

@mookid8000 just a quick question, what made you decide to make it configurable on the bus instead of configurable per Saga by letting it inherit an interface for instance? Not judging, just curious.

@MrMDavidson I have looked at a redis implementation for the lockprovider and locks but right now I am going with an implementation based on Azure Blob leases. 

We are running rebus instances on webjobs which are pretty scalable too. We use it to achieve eventual consistency for Cosmos Db and video processing for example. When we apply some kind of autoscaling principle, we have the necessity of distributed locking. So for now it looks like the EnforceExclusiveAccess only gets usable when we can provide a locking mechanism ourselves.

mookid8000:
> what made you decide to make it configurable on the bus instead of configurable per Saga by letting it inherit an interface for instance? Not judging, just curious.

Mostly the fact that I just didn't consider the other option 😉 and because @MrMDavidson 's implementation was a huge inspiration and he did it like this 😁

mookid8000:
> an implementation based on Azure Blob leases.

that's actually pretty brilliant 👍

MrMDavidson:
@cldons 
> I am going with an implementation based on Azure Blob leases.

Very nice idea! I didn't even think of such an implementation. That would have worked for us, too! Ended up using Redis based locks just because that's a pattern we use in a few other places on our platform.

> We use it to achieve eventual consistency for Cosmos Db and video processing

Sounds like we have a similar workload. We're doing audio processing and production.

@mookid8000 
>Mostly the fact that I just didn't consider the other option 😉 and because @MrMDavidson 's implementation was a huge inspiration and he did it like this 😁

Just a clarification; in Rebus.SingleAccessSagas you enable it on the bus (`config.Options(o => o.EnableSingleAccessSagas());`) but all that does is install the `IIncomingStep` which hooks into the Rebus pipeline. Once enabled it's actually opt in by virtue of one, or more, handlers being sagas **and** being tagged with `ISingleAccessSaga`. This way the overhead is minimal (some type introspection, and that's it) for most messages and sagas. And it's only a subset of our sagas which are both CPU intensive and fairly chatty that acquire any locks / face potential deferral of messages.

PR #647, opened 2017-10-13
Update README.md

LindaLawton:
example above uses someContainerAdapter not myFavoriteIocContainer.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Configure.With(someContainerAdapter)
 // have IBus injected in application services for the duration of the application lifetime    
 
 // let the container dispose the bus when your application exits
-myFavoriteIocContainer.Dispose();
+someContainerAdapter.Dispose();
 ```
 
 which will stuff the resulting `IBus` in the container as a singleton and use the container to look up


README.md
```diff
@@ -82,7 +82,7 @@ Configure.With(someContainerAdapter)
 // have IBus injected in application services for the duration of the application lifetime    
 
 // let the container dispose the bus when your application exits
-myFavoriteIocContainer.Dispose();
+someContainerAdapter.Dispose();
```
mookid8000: Hi Linda,

Thanks for the PR 😄 

I understand why this can be unclear at first, but it is actually not an error.

The usage pattern is basically like this:
```csharp
using(var container = new YourIoCContainer())
{
    Configure.With(new TheContainerAdapter(container))
        .(...)
        .Start();

    // application runs now
    Thread.Sleep(10000);

    // container disposes its contents, including the bus
}
```
which means that the container adapter is simply used to adapt it to be used by Rebus to look up message handlers

mookid8000:
Thanks for the PR, but I am closing it as per the comment in the source code above.... thanks anyway 😄

README.md
LindaLawton: I am just learning this seams confusing to me.    But i will give it a go :)

mookid8000: cool!

The thing is: IoC containers (at least those that understand how they ought to work 😉 ) are good at functioning as a sort of "application container", i.e. they can contain everything that it takes to make out an app, and then they can dispose everything that needs to be disposed, when the container itself is disposed.

Therefore it makes sense to use an IoC container in conjunction with Rebus by carrying out the following steps:

1. Create IoC container
1. Configure Rebus to use IoC container
1. Let application run
1. Dispose IoC container when the application shuts down

I hope that it makes sense now 😄

PR #650, opened 2017-10-25
Allow the bus created in SagaFixture to be configured externally

boyanio:
When using custom serializer for messages, it could be difficult to write unit tests using _SagaFixture_. This PR allows a configurer to be passed externally and it is overridden in the _SagaFixture_ with the minimum configurations the fixture needs.


benne:
Oh, yes - we need this.

mookid8000:
excellent – thanks 😄 

it'll be out in 4.1.0 in a few minutes

Issue #401, opened 2016-01-29
Make saga data serializer configurable


mookid8000:
and default to JSON.NET as usual, but allow for customizing this aspect too

mookid8000:
If the saga data serializer could be configured, it would also be trivial to implement compression and [encryption](http://stackoverflow.com/questions/34986605/encrypt-saga-data-in-rebus) as serializer decorators.

mookid8000:
Change persisters to use raw bytes as their storage format - preferably in a way that is backwards-compatible and doesn't break when running on data stored in the old format:

- [ ] `InMemorySagaStorage`
- [ ] `PostgreSqlSagaStorage`
- [ ] `SqlServerSagaStorage`
- [ ] `MongoDbSagaStorage`
- [ ] `RavenDbSagaStorage`

jeffreyabecker:
One thing to note: If the AzureStorage module is accepted it will need the serializer to expose the content type of the resulting data and the appropriate text encoding if applicable.

mookid8000:
Is this because you are filling out the `Content-Type` property of the blob?

Another way to do this would be similar to how the existing message serializer works - it transforms tuples of `(headers, bytes)` to/from tuples of `(headers, object)`, and in that process it can add/remove headers as needed.

The Azure Storage-based saga persister could pass a tuple of (properties, ISagaData) to the saga data serializer, which would then return (properties, bytes).

This way, the requirements of this particular blob storage will not leak into the surface API of the saga data serializer (which does not even exist yet 😉 )

mookid8000:
The saga data serializer might be configurable, but it has to be an option provided by the chosen saga data persister.

E.g. if MongoDB is chosen for persisting saga data, the data will be serialized using the Mongo driver's built-in BSON serializer.

With Postgres, the saga data should probably be saved as a JSON string into a proper JSON column.

And in future SQL Server versions it might become interesting to do something similar.

I'm closing this issue for now.

PR #651, opened 2017-11-05
Allow saga data to be serialized with custom serializer

boyanio:
These changes allow saga data serialization to be controlled externally, like message serialization. There are two scenarios, in my head, in which this is useful:

## Customize existing JSON serialization

There are situations where one would want to customize the default JSON serialization behavior, f.x., by injecting custom converters. This can be achieved by doing something like this:

```csharp
Action<JsonSerializerSettings> jsonSettingsConfigurer = settings =>
{
  settings.Converters.Add(new MyCustomConverter());
};

Configure.With(...)
  .Sagas(s => s.StoreInMemory(new JsonSagaDataSerializer(jsonSettingsConfigurer)))
...
```

## Use other type of serialization

There could be situations where one would use another type of serialization, f.x., XML. One could write a custom implementation of `ISagaDataSerializer` and inject it in the desired storage mechanism.

```csharp
Configure.With(...)
  .Sagas(s => s.StoreInMemory(new XmlSagaDataSerializer()))
...
```

## Backward compatibility

The changes are supposed to be 100% backward-compatible with the previous implementation, i.e. if one updates Rebus, everything should work as expected. The only difference is that `FileSystemSagaSnapshotStorage` and `LoggerSagaSnapperShotter` will include the C# types in the JSON output (with a `$type` property). This is a side effect of the default `JsonSagaDataSerializer`, because it always outputs `$type` properties.

## Future intentions

If this PR gets accepted, I plan to extend Rebus.SqlServer with the possibility to customize the saga data serialization.

---


mookid8000:
Hi @boyanio – nice work, although I do have some concerns...... I hope you will be patient and discuss them with me 😄 

Issue #401 "Make saga data serializer configurable" was about the exact same thing, and after thinking about it for a while, I realized that there was no good, clean way of making the saga data serializer _generally_ configurable – the reason being, that some saga persisters (e.g. RavenDB and MongoDB) have native support for object persistence and thus are capable of completely bypassing the need for at serializer (at least from Rebus' perspective).

Therefore, the conclusion in #401 went like [this](https://github.com/rebus-org/Rebus/issues/401#issuecomment-255481584):

> The saga data serializer might be configurable, but it has to be an option provided by the chosen saga data persister. (...)

which means that a configurable saga data serializer should be an option in e.g. Rebus.SqlServer, provided for the SQL Server saga persister only.

In addition to this, the saga data snapshots should NOT be created using the same serializer as the saga data serializer, because the snapshots are not meant to be used for roundtripping saga data and thus can work with much less fuss than the proper saga data serializer.

Moreover – if e.g. the SQL Server saga persister could some time be configured to use a binary serialization for saga data – it would suddenly make saga data snapshots pretty hard to read 😄 

Please let me know your thoughts.

boyanio:
Hi @mookid8000 I see your point. I have to say sorry for not doing my research before, but now I got a bit better picture of the situation.

Let me start with some background why I think this is useful. In our set-up, we use XML rather than JSON. We have a lot of domain objects that implement `IXmlSerializable` (and in fact there is a `JsonConverter` implementation for them). For the messages we use XML serialization and we want to reuse our models (DTOs) in the saga data as well, otherwise we have to do write wrapping/unwrapping logic constantly. So in my scenario, even if we can't use XML, configuring the JSON serialization (injecting custom converters) will do us fine in a way.

I totally see the point that saga data serialization should be dependent on the chosen storage mechanism and different for snapshots. My current implementation supports configuring a separate serializer for each case, using the configuration extensions. The default one is JSON for all, as before. Maybe `ISagaDataSerializer` could be split into two interfaces:
- `ISagaDataSerializer` that serializes to/from byte array for storage only, where needed. Actually I started with this version, but then kinda tried to fit it all :-)
- `ISagaDataTextualVisualizer` (or similar) to serialize saga data to string (i.e. one-way only) for snapshots and logging

In any case the suggested changes allow manually to configure the serializer for each storage mechanism - be it InMem, FileSystem or snapshot (I encountered one configurer I have forgotten to extend to allow configuring the serializer, though).

Is your concern that `ISagaDataSerializer` lies in the core project and creates expectation that it must be used by any storage provider?

mookid8000:
Hi @boyanio , this is just to say that I did not forget this PR 🙂 I just need some time to process it, and that time comes around Thursday this week

mookid8000:
...and by "Thursday this week" I mean "some time later" – you're on my Trello board, I will not forget you, and I may be closing in on an idea, (slow-)cooking in the back of my mind, just the way it's supposed to be...

mookid8000:
> (....)  I will not forget you (...)
– Mogens

🤣

mookid8000:
I guess that slow-cooking never materialized into anything. Sorry about that.

The problem (at least as I see it right now), is that whether a serializer is necessary depends a lot on the chosen persistence. E.g. with MongoDB and RavenDB, one would probably prefer to have the respective drivers do the saving, preserving everything as rich documents.

PR #658, opened 2017-11-10
FileSystemTransport now correctly supports OneWayClient

pheiberg:
The FileSystemTransport has a UseFileSystemAsOneWayClient extension method, but it's currently not working, since it does not update the configuration correctly. This pull request adds the vital missing piece.

### DIFF
--- a/Rebus/Transport/FileSystem/FileSystemTransportConfigurationExtensions.cs
+++ b/Rebus/Transport/FileSystem/FileSystemTransportConfigurationExtensions.cs
@@ -38,6 +38,7 @@ namespace Rebus.Transport.FileSystem
             if (baseDirectory == null) throw new ArgumentNullException(nameof(baseDirectory));
 
             configurer.Register(context => new FileSystemTransport(baseDirectory, null));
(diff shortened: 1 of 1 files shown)


Rebus/Transport/FileSystem/FileSystemTransportConfigurationExtensions.cs
```diff
@@ -38,6 +38,7 @@ public static void UseFileSystemAsOneWayClient(this StandardConfigurer<ITranspor
             if (baseDirectory == null) throw new ArgumentNullException(nameof(baseDirectory));
 
             configurer.Register(context => new FileSystemTransport(baseDirectory, null));
+            OneWayClientBackdoor.ConfigureOneWayClient(configurer);
```
mookid8000: looks perfekt! 👍 🎉

mookid8000:
The problem was that a one-way client using the file system transport would still end up trying to receive messages... but since it was not configured with an input queue name, this would cause `Path.Combine(path1, path2)` to fail with an `ArgumentNullException` each time the worker tried to receive the next message.

Issue #657, opened 2017-11-09
Add ability to fail fast on certain exception types


mookid8000:
Rebus 1 had the ability to have the number of delivery attempts configured for individual exception types, such that e.g. `ConcurrencyException` would retry 10 times and `ArgumentException` would not retry at all.

This was o.k. in some ways, but it raised a bunch of annoying questions like e.g. "what to do with different exceptions", etc.

Now that I have been asked "can Rebus retry a varying number of times, depending on the exception type?" 100 times, and each and every time it turned out to be because it doesn't make sense to retry at all for some specific exception types – so in fact it wasn't about configuring the number of times, it was more about the ability to fail fast on specific exception types.

Therefore: Add a pipeline step that disables retries for specific exception types, which can be enabled e.g. like so:
```csharp
.Options(o => {
    o.DisableRetriesFor<DomainException>();
});
```
or something like that.

(am unsure about the syntax for disabling the retries.... it's a little bit ugly that it's directly on the `OptionsConfigurer`, but I don't know where else to put it....)

boyanio:
WCF has `FaultException<>` to wrap some faults from the service, so one could introduce a general exception wrapper, f.x. `FailFastException<TException>`, that automatically fails fast.

Furthermore, to allow custom logic for some other exceptions that should fail fast, one could introduce a simple interface, f.x. `IFailFastExceptionHandler` with one method `bool ShouldFailFast(Exception e)` to determine if an exception should lead to fail-fast or retry multiple times. This approach, I believe, is more flexible as it allows you to check not only on the exception type, but also on the exception data.

mookid8000:
@boyanio I like the `FailFastException<>` approach! And I also like what you said about the `IFailFastExceptionHandler`.

I am thinking that the cleanest way of making this work, would be to roll back everything (like when an ordinary exception is caught), and then tell the `IErrorTracker` that the message has reached the limit for how many times it is to be attempted to be consumed.

This would make the `SimpleRetryStrategyStep` forward the message to the error queue when it is received shortly afterwards.

This reason it requires these two steps, is if the user has a unit of work with e.g. database transactions etc., then it is important that these things are rolled back properly – therefore, it is not possible to simply forward the message in the same transaction context.

boyanio:
@mookid8000 I see that `SimpleRetryStrategyStep` gets registered via `IRetryStrategy`, which means theoretically that I, as a user, can define my own retry strategy. So there are two ways to do it conceptually:

[...]

PR #664, opened 2017-11-28
New custom step to allow failed messages to go "directly" to error queue

boyanio:
A new incoming step `FailFastStep` checks if a message has been marked as failed with a specific exception, it sends the message directly to the error queue without retrying further. The default check happens for a custom exception of type `FailFastException`. Custom logic could be built by implementing `IFailFastChecker`.

This step uses only `IErrorTracker` and relies on its semantics (f.x., that calling `RegisterError` becomes closer to `HasFailedTooManyTimes`). This step does not take into consideration transactions as they are supposed to be abroted by `SimpleRetryStrategyStep` when the first exception has been caught. By marking the message as "failed too many times", the `SimpleRetryStrategyStep` will automatically move the message to the configured error queue.

This PR comes as a possible solution to #657

---

### DIFF
--- a/Rebus.Tests/Integration/TestFailFast.cs
+++ b/Rebus.Tests/Integration/TestFailFast.cs
@@ -1,0 +1,145 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using Rebus.Activation;
+using Rebus.Bus;
+using Rebus.Config;
+using Rebus.Exceptions;
+using Rebus.Extensions;
+using Rebus.Logging;
+using Rebus.Messages;
+using Rebus.Retry.Simple;
+using Rebus.Routing.TypeBased;
+using Rebus.Tests.Contracts;
+using Rebus.Tests.Contracts.Extensions;
+using Rebus.Transport.InMem;
+using Rebus.Retry.FailFast;
+
+#pragma warning disable 1998
+
+namespace Rebus.Tests.Integration
+{
+    [TestFixture]
+    public class TestFailFast : FixtureBase
+    {
+        static readonly string InputQueueName = TestConfig.GetName($"test.rebus2.retries.input@{GetMachineName()}");
+        static readonly string ErrorQueueName = TestConfig.GetName("rebus2.error");
+
+        BuiltinHandlerActivator _handlerActivator;
+        IBus _bus;
+        InMemNetwork _network;
+
+        void InitializeBus(int numberOfRetries, IFailFastChecker failFastChecker = null)
+        {
+            _network = new InMemNetwork();
+
+            _handlerActivator = new BuiltinHandlerActivator();
+
+            _bus = Configure.With(_handlerActivator)
+                .Logging(l => l.Console(minLevel: LogLevel.Warn))
+                .Transport(t => t.UseInMemoryTransport(_network, InputQueueName))
+                .Routing(r => r.TypeBased().Map<string>(InputQueueName))
+                .Options(o =>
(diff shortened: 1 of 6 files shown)


mookid8000:
Awesome work 😄

mookid8000:
I looked at the code, and I noticed how the `FailFastStep` would be doing a fair amount of work, even in the trivial case where no exceptions were thrown. 

I had a suspicion that that would have a negative impact on dispatch performance, so I ran Rebus' `TestDispatchPerformance` test with the step disabled and enabled respectively – these were the results:

```
Initial:
Avg msg/s: 19765

With original FailFastStep:
Avg msg/s: 16551
```

So, the rate changed from 19k msg/s to 16k msg/s.

Then I changed the code to work, only kicking in inside the `catch` block of `FailFastStep` (which also required some other changes) – it gave me this:

```
Avg msg/s: 19021
```

keeping the rate around 19k msg/s, which is acceptable.



(*) Please note that the actual numbers listed here are artificial and only meant to be used to compare the impact of changes like this. The code is "instrumented" by interleaving profiling steps with the original pipeline steps, allowing for generating interesting information like this (from the last run):
```

Running 10 samples with 100000 msgs and mode DefaultNew
Performing sample 1: 5,92529
Performing sample 2: 4,85979
Performing sample 3: 4,7827
Performing sample 4: 5,05213
Performing sample 5: 4,90567
Performing sample 6: 5,24323
Performing sample 7: 4,93766
Performing sample 8: 4,86724
Performing sample 9: 4,93434
Performing sample 10: 4,9793
10 runs
Avg s: 5,04873
Avg msg/s: 19807

Med s: 4,93766
Med msg/s: 20253

Pipeline invoker: DefaultNew

Stats:
    Rebus.Pipeline.Receive.DispatchIncomingMessageStep: 00:00:08.8240427 (8,3 %)
    Rebus.Retry.Simple.SimpleRetryStrategyStep: 00:00:04.8789620 (4,6 %)
    Rebus.Retry.FailFast.FailFastStep: 00:00:34.1794105 (32,1 %)
    Rebus.Pipeline.Receive.HandleRoutingSlipsStep: 00:00:02.2214585 (2,1 %)
    Rebus.Pipeline.Receive.HandleDeferredMessagesStep: 00:00:03.0630926 (2,9 %)
    Rebus.Sagas.LoadSagaDataStep: 00:00:01.4088631 (1,3 %)
    Rebus.Pipeline.Receive.DeserializeIncomingMessageStep: 00:00:33.5610565 (31,5 %)
    Rebus.Pipeline.Receive.ActivateHandlersStep: 00:00:18.4802259 (17,3 %)
```

mookid8000:
it's out in 4.2.0-b01 now 😄

boyanio:
Good point! I hadn't really thought about performance in this case. An alternative solution could be to "merge" `FailFastStep` with `SimpleRetryStrategyStep`. Then one could apply fail-fast logic directly in the `catch` clause. But then we can discuss whether this should be part of `SimpleRetryStrategyStep`'s job or the step gets to do too much.

Another possible solution could be to introduce a new service, f.x.

```
interface IFailFastHandler
{
  bool HasAnyExceptions(string messageId);
  IEnumerable<Exception> GetExceptions(string messageId);
  void MarkAsFailFast(string messageId, Exception exception);
}
```

Then one could implement this service on `InMemErrorTracker` to provide "internal" way of checking if a message is a candidate for fail-fast and of marking a message as fail-fast. As I believe the logic of getting all exceptions every time is slowing the step (apart from the mere fact of introducing a new step).

PR #700, opened 2018-05-22
Made storage for saga exclusive lock configurable by using an interface

torangel:
I have tried two alternative approaches for implementing "centralized-saga-locking" by storing locks in Redis. Locking in-process (as Rebus already supports) is not an option as we are running multiple workers.
The first solution is documented here https://github.com/idfy-io/rebus.redissagalock

The second one is documented in this pull request which basically introduces a simple interface for configuring where locks are stored. It would then be easy to implement this interface and store the locks in redis or some other store.

Feel free to comment or propose alternative approaches.

---

### DIFF
--- a/Rebus.Tests/Assumptions/TestLockContention.cs
+++ b/Rebus.Tests/Assumptions/TestLockContention.cs
@@ -80,13 +80,14 @@ Elapsed: {elapsed.TotalSeconds:0.0}
 
         static IEnumerable<IIncomingStep> GetSteps()
         {
-            yield return new EnforceExclusiveSagaAccessIncomingStep();
+            var handler = new HandleSagaExlusiveLockInConcurrentDictionary();
+            yield return new EnforceExclusiveSagaAccessIncomingStep(handler);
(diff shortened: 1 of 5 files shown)


mookid8000:
Great work 👍 it'll be out in a 5 prerelease in a few minutes... brb

mookid8000:
I made a few changes – first off, I renamed `IHandleSagaExlusiveLock` to  `IExclusiveSagaAccessLock`, because I thought that the `IHandle***` sounded a little bit like Rebus' message handlers.

Then I brought back the `CancellationToken` support in the `AquireLockAsync` method, because a distributed lock could be implemented in a way that waits a while before giving up on its attempt to acquire the lock, and that would be a bummer if Rebus was shutting down.

Other than that, I think it's really great work 😄 it's super to have this capability added to Rebus! Thanks!

mookid8000:
It's out as 5.0.0-b11 now

torangel:
It is a pleasure contributing to such a excellent piece of software as Rebus. Keep up the good work you (and others) are doing!

PR #722, opened 2018-07-20
Add WillBeInvoked property to HandlerInvoker

### LINKED ISSUE #713: Question: API for assessing if a HandlerInvoker will be skipped?
We have some custom `IIncomingStep`s in our pipeline. One of the steps, early in the pipeline, may call `SkipInvocation()` on one or more of the `HandlerInvoker`s for the message. One thing I noticed in the course of doing this is that if the invoker being skipped was an `IAmInitiatedBy<TMessageType>` then when the pipeline gets to `LoadSagaDataStep` the saga data will still be initialised and persisted to the saga store. This means that if another `TMessageType` with the same correlation comes along but **does** allow the `IAmInitiatedBy<TMessageType>` to be invoked then the handler's checks of `IsNew` to preserve idempotency would return false. Whilst we're definitely doing some esoteric stuff it does seem odd to me that there is no corresponding API to `SkipInvocation()` for testing if the invoker will actually be called. Eg. I can imagine https://github.com/rebus-org/Rebus/blob/master/Rebus/Sagas/LoadSagaDataStep.cs#L62 being modified to;
```C#
var handlerInvokersForSagas = context.Load<HandlerInvokers>()
  .Where(l => l.WillExecute)
  .Where(l => l.HasSaga)
  .ToList();
```

Which would save a potential network round trip to the data store.

Ideally the ability to check if the invoker will actually be called would be public.

Any thoughts on this?

MrMDavidson:
Add a `WillBeInvoked` property to `HandlerInvoker` and `HandlerInvoker<T>`. This is set to false if `SkipInvocation()` has been called on an invoker. This allows `IIncomingStep`, and other infrastructure, to act accordingly if an invoker is getting skipped.

Resolves #713 .

---

### DIFF
--- a/Rebus/Pipeline/Receive/HandlerInvoker.cs
+++ b/Rebus/Pipeline/Receive/HandlerInvoker.cs
@@ -78,6 +78,11 @@ namespace Rebus.Pipeline.Receive
         /// not depend on it!)
         /// </summary>
         public abstract object Handler { get; }
+
+		/// <summary>
+		/// <c>true</c> if the handler will be invoked, as per normal, <c>false</c> if <see cref="SkipInvocation"/> has been called or the invoke will otherwise be skipped
+		/// </summary>
+		public abstract bool WillBeInvoked { get; }
     }
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


mookid8000:
cool 👍

mookid8000:
it'll be out as soon as I get near my computer 😁

MrMDavidson:
Cheers. Happened to have a moment of spare time on a quiet Friday arvo so figured I'd quickly do it.

Issue #727, opened 2018-08-27
ThreadPoolWorker uses Thread.Sleep (via ISyncBackoffStrategy) as part of async method, blocking a thread pool thread


Plasma:
Hello,

## Symptoms
We noticed when writing our own ``ITransport`` that made heavy use of async/await, that we were receiving random multi-second delays processing messages, throughput became very low, and async callbacks were taking multiple seconds longer than expected in our internal code.

## Root Cause
``ThreadPoolWorker`` has a method ``TryAsyncReceive`` which then awaits ``ReceiveTransportMessage`` (that calls ``ITransport.Receive`` async). See https://github.com/rebus-org/Rebus/blob/master/Rebus/Workers/ThreadPoolBased/ThreadPoolWorker.cs#L101

When the async context resumes after ``ReceiveTransportMessage`` returns, we are likely in a thread-pool thread.

We then call ``WaitNoMessage`` (https://github.com/rebus-org/Rebus/blob/master/Rebus/Workers/ThreadPoolBased/ThreadPoolWorker.cs#L113) when there is no message, which ends up calling ``Thread.Sleep`` via ``ISyncBackoffStrategy`` https://github.com/rebus-org/Rebus/blob/master/Rebus/Workers/ThreadPoolBased/DefaultSyncBackoffStrategy.cs#L69

This blocks the thread-pool thread for the backoff time, which could be multiple seconds.

This in turn causes other async tasks queued up on that thread pool thread to be delayed from being executed/resumed for at least this amount of time.

## Proposed Fix
I can submit a PR that will rename ``ISyncBackoffStrategy`` to ``IAsyncBackoffStrategy`` and make all methods ``Task`` return instead of ``void``, and replace the ``Thread.Sleep`` with ``Task.Delay`` instead.

This would unblock the thread pool thread and avoid pausing/slowing down other async tasks unrelated to Rebus.

Let me know if that's the appropriate change in a PR you would like, or if you have any questions. 

Thanks for a great library.

mookid8000:
Great diagnosis, and probably also a pretty great solution.... I cannot readily see why your solution should not work, but I have this eerie feeling that there's a reason why the back-off strategy blocks the THREAD and does not simply `await Task.Delay(..)`.

You're most welcome to try and change it, and then we can release it in a 5 betaversion and try it out. 😄

Plasma:
[...]

PR #728, opened 2018-08-27
Replace ISyncBackoffStrategy with IAsyncBackoffStrategy as its used from async context

Plasma:
(Regarding https://github.com/rebus-org/Rebus/issues/727)

This PR replaces the ``ISyncBackoffStrategy``, which is only used by ``ThreadPoolWorker``, to now be ``IAsyncBackoffStrategy`` and use ``await Task.Delay`` instead of ``Thread.Sleep`` to avoid blocking thread pool threads, as the calling code is async.

---

### DIFF
--- a/Rebus.Tests/Backoff/TestBackoffBehaviorWhenBusy.cs
+++ b/Rebus.Tests/Backoff/TestBackoffBehaviorWhenBusy.cs
@@ -43,10 +43,10 @@ namespace Rebus.Tests.Backoff
                     o.SetBackoffTimes(TimeSpan.FromSeconds(0.2));
 
                     // install the snitch
-                    o.Decorate<ISyncBackoffStrategy>(c =>
+                    o.Decorate<IAsyncBackoffStrategy>(c =>
                     {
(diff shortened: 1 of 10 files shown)


Plasma:
@mookid8000 I note that https://github.com/rebus-org/Rebus/blob/bcbb00c0b46b705d18b70018b73ef38d9580af80/Rebus/Workers/ThreadPoolBased/ThreadPoolWorker.cs#L64 is async and not awaited, so no delay is performed here.

Perhaps we need a mix of Wait/WaitAsync WaitError/WaitErrorAsync etc?

mookid8000:
ah – I changed it [here](https://github.com/rebus-org/Rebus/blob/master/Rebus/Workers/ThreadPoolBased/ThreadPoolWorker.cs#L147), but I missed that other one....

and yes, they should probably both come in sync/async variants, depending on whether the calling code needs a delay before resuming execution or it needs to actually pause the thread

Plasma:
I can submit a PR if you like with sync/async variants (and perhaps rename the interface to just IBackoffStrategy)

mookid8000:
great idea 👍

PR #731, opened 2018-08-31
Include portable-pdbs in nuget package and enable SourceLink

### LINKED ISSUE #730: Symbol packages?
Are there symbols available anywhere for the NuGet packages?

If not could they be included for builds in the future? There's some information about how to do that at

* https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages
* http://www.symbolsource.org/Public/Wiki/Publishing

poizan42:
This uses the new [SourceLink v2](https://github.com/dotnet/sourcelink/) support available in Visual Studio 2017 15.3+. pdb files are included in the nuget package [as recommended by Microsoft](
https://github.com/aspnet/Universe/issues/131#issuecomment-363269268).

Fixes #730

---

### DIFF
--- a/Rebus/Rebus.csproj
+++ b/Rebus/Rebus.csproj
@@ -17,6 +17,17 @@
     <RepositoryUrl>https://github.com/rebus-org/Rebus</RepositoryUrl>
     <RepositoryType>git</RepositoryType>
     <PostBuildEvent></PostBuildEvent>
+    <Product>Rebus</Product>
+    <PackageId>Rebus</PackageId>
+    
+    <!-- Optional: Publish the repository URL in the built .nupkg (in the NuSpec <Repository> element) -->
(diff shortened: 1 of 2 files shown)


poizan42:
So this won't work with older Visual Studio versions, the alternative is to use "non-portable" pdbs and GitLink, but that is more complicated to set up and ultimately deprecated anyways.

poizan42:
Uhh it breaks CI, seems that it doesn't do a regular git checkout.

poizan42:
So I disabled SourceLink when building from CI

mookid8000:
coool 👍

mookid8000:
Rebus 5.0.0-b18 has PDBs in the NuGet package now, and I just pushed it to NuGet.org.

Thanks for fixing this!

PR #736, opened 2018-09-18
Removed .ConfigureAwait(false) calls

### LINKED ISSUE #726: HttpContext.Current not available in outgoing message pipeline
Rebus 4.2 switched to using `.ConfigureAwait(false)` in async calls (see commit f1a50e94f1a1624512e83fbcc62464d767ef3ac1). This causes HttpContext.Current to be not available within outgoing steps of the outgoing message pipeline because HttpContext.Current unfortunately does not flow across async calls using the logical call context or AsyncLocals but a special AspNetSynchronizationContext. Obviously, the same issue may happen for anything else which flows by using a synchronization context.

While it is in general advisable for library authors to use `.ConfigureAwait(false)` within their own library if their own code does not depend on any synchronization context, this does not hold true anymore if there are extension points where non-library code may be executed as well.

Please note that this issue is very subtle. We had Rebus configured with type based routing and subscriptions stored centrally in SQL Server. On send operations HttpContext.Current was available in outgoing steps because the type based routing effectively runs synchronously albeit the interfaces being asynchronous. On publish operations on the other hand HttpContext.Current was not available anymore in outgoing steps because the destination addresses where looked up in SQL Server asynchronously first, causing the thread to change.

We were able to store HttpContext.Current within an AsyncLocal prior to invoking Rebus and accessing the HttpContext from there, however it would be nice if we would not need to resort to such workarounds.

We suggest to not use `.ConfigureAwait(false)` across the entire library but only at places where no non-library code may be executed afterwards in order to mitigate such issues.

oliverhanappi:
Fixes #726 
---

### DIFF
--- a/Rebus.Tests/Workers/ThreadWorkerIntegrationTests.cs
+++ b/Rebus.Tests/Workers/ThreadWorkerIntegrationTests.cs
@@ -45,13 +45,11 @@ namespace Rebus.Tests.Workers
             {
                 startedHandle.Set();
 
-                await Task.Delay(TimeSpan.FromSeconds(10))
-                    .ConfigureAwait(false);
+                await Task.Delay(TimeSpan.FromSeconds(10));
             });
 
             // act
-            await _bus.Send("something")
-                .ConfigureAwait(false);
+            await _bus.Send("something");
 
             startedHandle.WaitOne();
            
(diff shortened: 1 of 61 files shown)


mookid8000:
thanks 👍 – it'll be out in Rebus 4.2.2 in a few minutes

PR #739, opened 2018-10-03
Posibility to create custom HandlerInvoker

Liero:
Added extension point to ActivateHandlerStep that makes it possible to create custom HandlerInvoker instaces without replacing entire step. Custom HandlerInvoker allows to add some logic before or after the message is invoked

---


mookid8000:
Cool, thanks 👍 it's out in Rebus 5.0.0-b20 in a few minutes

Issue #740, opened 2018-10-09
`RebusTimeMachine` does not seem to use the `incrementSlightlyOnEachInvocation` argument


boyanio:
Hey,

It has been a while since I noticed that `RebusTimeMachine.FakeIt` does not seem to use the `incrementSlightlyOnEachInvocation` argument. I suppose the correct implementation should be:

```
public static void FakeIt(DateTimeOffset fakeTime, bool incrementSlightlyOnEachInvocation = true)
{
  RebusTime.CurrentTimeFactory = () =>
  {
    var timeToReturn = fakeTime;
    if (incrementSlightlyOnEachInvocation)
    {
      fakeTime = fakeTime.AddTicks(1);
    }
    return timeToReturn;
  };
}
```

Do you agree? :-)

mookid8000:
I agree much 😁

PR #742, opened 2018-10-09
Make an actual use of the argument `incrementSlightlyOnEachInvocation`

boyanio:
In unit tests today one can call `RebusTimeMachine.FakeIt` to fake the current time in Rebus. The method creates a factory that increments the ticks every time `RebusTime.Now` is called. In some scenarios this is not wanted, so there exists the argument `incrementSlightlyOnEachInvocation` which can be set to `false`. Unfortunately, the argument is not used inside the created factory. See [issue](https://github.com/rebus-org/Rebus/issues/740) for more details.

---


mookid8000:
Good catch 😁thanks!

Issue #747, opened 2018-10-28
Question about Routes vs Transport


dicko2:
I'm working on a Kafka transport, and i need to subscribe to all the topics on startup.

as far as i can see the TypeBasedRouter does not expose the full list of topics in anyway, so i think that this will require a new method to return the full list from _messageTypeAddresses.

Is that right or am I missing something?

I am thinking I'll need something like the below in the configuration

```csharp

        public static void UseKafka(this StandardConfigurer<ITransport> configurer,
            string brokerList, string groupId)
        {
            
            configurer.Register(c =>
            {
                var router = c.Get<IRouter>();
                var rebusLoggerFactory = c.Get<IRebusLoggerFactory>();
                var asyncTaskFactory = c.Get<IAsyncTaskFactory>();
                var listOfTopics = router.GetListOfTopics();
                return new KafkaTransport(rebusLoggerFactory, asyncTaskFactory, listOfTopics, brokerList, groupId);
            });
        }
```

mookid8000:
> (...) and i need to subscribe to all the topics on startup

Why? 😄 I'm not an expert on Kafka, but to me it would seem like all endpoints would need to subscribe to these topics:
- a topic for the input queue
- a topic for each topic explicitly subscribed to via `await bus.Subscribe<T>(...)` (which in turn translates into `await bus.Advanced.Topics.Subscribe(someString)`)

This way, `await bus.Send(...)` (which always sends to 1 recipient) would append to some particular endpoint's input queue topic, and `await bus.Publish(...)` would append to the event's topic.

As far as I can tell, there's no need to subscribe to ALL topics for anyone..... or am I missing something here?

dicko2:
Perhaps I'm miss understanding the use. I've just been looking at the samples. 

I guess there is an architectural question as to if you use the topic mechanism of the underlying transport, or house this in the abstraction. I am trying to use the topic mechanism of the messaging platform. 

Following the example code, I pass a func to the handler for type X, but then do I also need to call subscribe for type X as well?

Also where does subscribe connect to the transport layer? As far as I can see the only connection is at instantiation. 

Methods are

Send
Receive 
Create queue 

And the error library is the only thing that calls create queue. So I think I am missing how the transport is told what topics to listen to. 

I've made public the code I'm working on if you want to have a look here:

Github.com/agoda-com/Rebus.kafka 

This is (a work in progress) using the version compiled from my Pr. 

The single test in the test project is me following the example code, and may explain where I am coming from better

[...]

PR #748, opened 2018-10-28
expose full list of routes from IRouter

dicko2:
In regards to #747 

This is the only way i can think to make it work

---


mookid8000:
I'm closing this PR – not because I don't want to accept it, but because I don't think it's necessary to be able to implement a Kafka transport (this in in relation to #747).

PR #759, opened 2018-11-19
Added count methods for InMemoryNetwork

CasperTDK:
We have a few small microservices that does not need to be durable. So we just use In Memory transport. However, we do want to query the status of the services and an insight into queue count is helpful.
---


mookid8000:
cool ,thanks 👍

CasperTDK:
@mookid8000  Thanks for the merge. When is a new release planned?

mookid8000:
Before you turn 30!

mookid8000:
(it's out in Rebus 5.0.2, which is on NuGet.org now.... thanks again 👍 )

PR #793, opened 2019-04-17
Make logging more symmetrical when sending and receiving messages

nativenolde:
@mookid8000 Hey mogens,

I've made logging more symmetrical when sending and receiving messages by using the same label.
This will change the original message from:
```
[DBG] Sending Testing.Library.CommandA -> "CRM"
[DBG] Sending Testing.Library.CommandB -> "ERP"
```

to this kind:
```
[DBG] Sending Testing.Library.CommandA, Testing, <ID> from "Whatever" -> "CRM"
[DBG] Sending Testing.Library.CommandB, Testing, <ID> from "Whatever" -> "ERP"
```

to better meet the other side:
```
[DBG] Dispatching Testing.Library.CommandA, Testing, <ID> to 1 handlers took 563 ms
[DBG] Dispatching Testing.Library.CommandA, Testing, <ID> to 1 handlers took 563 ms
```

This change brings the outgoing message id into the (debug) log, which is currently missing.
Especially if you're using Request/Reply using Rebus.Async, this helps to track the flow.

Sincerely,
nolde

---

### DIFF
--- a/Rebus/Pipeline/Receive/DispatchIncomingMessageStep.cs
+++ b/Rebus/Pipeline/Receive/DispatchIncomingMessageStep.cs
@@ -41,18 +41,17 @@ If no invokers were found, a RebusApplicationException is thrown.")]
         /// </summary>
         public async Task Process(IncomingStepContext context, Func<Task> next)
         {
-            var invokers = context.Load<HandlerInvokers>();
             var handlersInvoked = 0;
+            var invokers = context.Load<HandlerInvokers>();
             var message = invokers.Message;
 
-            var messageId = message.GetMessageId();
-            var messageType = message.GetMessageType();
+            var messageLabel = message.GetMessageLabel();
 
             // if dispatch has already been aborted (e.g. in a transport message filter or something else that
             // was run before us....) bail out here:
             if (context.Load<bool>(AbortDispatchContextKey))
             {
-                _log.Debug("Skipping dispatch of message {messageType} {messageId}", messageType, messageId);
+                _log.Debug("Skipping dispatch of message {messageLabel}", messageLabel);
                 await next();
                 return;
             }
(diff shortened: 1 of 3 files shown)


mookid8000:
hey @nativenolde , I think this is a really good initiative! There's a little problem, though, which I'll point out with a little code review discussion... just a sec

Rebus/Pipeline/Receive/DispatchIncomingMessageStep.cs
```diff
@@ -41,18 +41,17 @@ public DispatchIncomingMessageStep(IRebusLoggerFactory rebusLoggerFactory)
...
 
-            var messageId = message.GetMessageId();
-            var messageType = message.GetMessageType();
+            var messageLabel = message.GetMessageLabel();
 
             // if dispatch has already been aborted (e.g. in a transport message filter or something else that
             // was run before us....) bail out here:
             if (context.Load<bool>(AbortDispatchContextKey))
             {
-                _log.Debug("Skipping dispatch of message {messageType} {messageId}", messageType, messageId);
```
mookid8000: With a structured logging library like [Serilog](https://github.com/serilog/serilog), this will log an event accompanied by the following data:
```
{
	"messageType": "SomeType, SomeAssembly",
	"messageId": "4393umessageID"
}
```
making it possible to filter and aggregate logs using these values further down the line (e.g. in Kibana, if you're using the ELK stack).

Rebus/Pipeline/Receive/DispatchIncomingMessageStep.cs
```diff
@@ -41,18 +41,17 @@ public DispatchIncomingMessageStep(IRebusLoggerFactory rebusLoggerFactory)
...
+                _log.Debug("Skipping dispatch of message {messageLabel}", messageLabel);
```
mookid8000: ....conversely, this will result in logging
```
{
	"messageLabel": "SomeType, SomeAssembly, 4393umessageID"
}
```
which  is pretty useless without further processing.

mookid8000:
As you can see, the log format strings have to contain the individual fields.

So, I would love to have more symmetry around what is logged and how it looks, when Rebus logs send & receive of messages, but it has to retain the proper fields in the log statements.

Could you maybe adapt your PR slightly to accommodate this? 😄

nativenolde:
You're absolutely right. Thats not a good thing if a structured logger gets used. I didn't take that into account... 

My intension was to log the message id while sending... therefore i thought it would be good to 
streamline that by using the message label, but it's definitely better to log that into separate fields.

Would you please be so kind to add the message id into log when sending?

Issue #816, opened 2019-07-24
Can IBus.SendLocal() accept IDictionary<string,string> header instead of Dictionary<string,string>?


hellfirehd:
Currently:
```csharp
public interface IBus : IDisposable
{
  Task SendLocal(object commandMessage, Dictionary<string, string> optionalHeaders = null);
  . . .
}
```

Is there a particular reason why it expects a concrete `Dictionary<string,string>` for headers?  It would be helpful for me if it used the interface throughout.

I'm happy to submit a pull request.

mookid8000:
Sure, that's a great idea. 😄 

For consistency, it should be changed in all places accepting optional headers, though:

* [`IBus`](https://github.com/rebus-org/Rebus/blob/master/Rebus/Bus/IBus.cs) 
* [`ITransportMessageApi`](https://github.com/rebus-org/Rebus/blob/master/Rebus/Bus/Advanced/ITransportMessageApi.cs)
* [`IRoutingApi`](https://github.com/rebus-org/Rebus/blob/master/Rebus/Bus/Advanced/IRoutingApi.cs)
* [`ISyncBus`](https://github.com/rebus-org/Rebus/blob/master/Rebus/Bus/Advanced/ISyncBus.cs)
* [`ITopicsApi`](https://github.com/rebus-org/Rebus/blob/master/Rebus/Bus/Advanced/ITopicsApi.cs)

Happy to accept as PR on master, so it will go into Rebus 6.

PR #817, opened 2019-07-25
Update interfaces to accept IDictionary<string,string> for optionalHeaders

hellfirehd:
Fixes:  #816

---

### DIFF
--- a/Rebus.Tests.Contracts/Activation/ContainerTests.cs
+++ b/Rebus.Tests.Contracts/Activation/ContainerTests.cs
@@ -221,15 +221,15 @@ has failed too many times)
 
             public void Dispose() => _bus.Dispose();
 
-            public Task SendLocal(object commandMessage, Dictionary<string, string> optionalHeaders = null) => _bus.SendLocal(commandMessage, optionalHeaders);
+            public Task SendLocal(object commandMessage, IDictionary<string, string> optionalHeaders = null) => _bus.SendLocal(commandMessage, optionalHeaders);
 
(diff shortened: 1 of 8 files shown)


hellfirehd:
I think this could be a breaking change as all implementors must be updated to accept the IDictionary<string,string> as well.

mookid8000:
> I think this could be a breaking change as all implementors must be updated to accept the IDictionary<string,string> as well.

You're absolutely right: This is a breaking change, which is why we're scheduling it for Rebus 6.

Since all existing callers must be passing a `Dictionary<string, string>`, it should not require any changes to client code. My guess is that 99% of client code is of this kind.

The last 1% are people, who have implemented `IBus`/`ISyncBus`/etc. themselves, e.g. as a decorator, or as a hand-rolled fake for testing, and those people need to do a Search/Replace to prepend the `I`.

**BONUS PRO VISUAL STUDIO PRODUCTIVITY TIP IN CASE ANYONE NEEDS IT**: Did you know Visual Studio has support for **multiple cursors**, just like Sublime Text? 

Check this out! 🤠
👇
👇
For example, if you need to replace `Dictionary<string, string>` all the way down in the file you're standing in, then you can start out by selecting the first occurrence, and then you can `CTRL+ALT+.` to select each next occurrence of that text, placing a cursor in each selection. When you've selected all of them, then you can edit all of them at the same time and add the missing `I`.

mookid8000:
Thanks for fixing this 😄 it's out in Rebus 6.0.0-b09 in a few minutes

PR #823, opened 2019-08-12
Optimize `TransactionContext`

jods4:
`TransactionContext` uses 4 concurrent queues internally. 
Each one is a heavy object, implementation includes at least two arrays of length 32, and around ten fields, in 32 bits we're close to 400 bytes per queue, so 1.5Kb per `TransactionContext`.
Manipulating them is not cheap either.

I replaced them with events (i.e. `MulticastDelegate`), which is `null` if empty, otherwise a linked list of delegates (target + method pointer). Much, much cheaper.

`add` and `remove` on `event` are thread-safe in C#, generated code uses a compare-and-exchange loop. Invoking is also thread safe (based on a snapshot at time of invocation).

**Note that the original code had a race condition** where an action could for example be added to `OnCommitted` and not throw a completed exception, nor be called. 
**I did not fix this race condition and it still exists.**
Easiest fix might be to make all flags volatile and set them before invoking the corresponding event.

I did not bother with emptying the events that will only be called once (i.e. all of them except Commit).

I also extracted the throwing code in `OnXXX` method. It makes those methods very likely to be inlined by JIT.

Finally I removed some async machinery where it was excessive.

---

### DIFF
--- a/Rebus/Transport/TransactionContext.cs
+++ b/Rebus/Transport/TransactionContext.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.Collections.Concurrent;
+using System.Runtime.CompilerServices;
+using System.Threading;
 using System.Threading.Tasks;
 // ReSharper disable SuggestBaseTypeForParameter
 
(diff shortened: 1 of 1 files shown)


mookid8000:
This is really good work! 👍 thanks! I'll publish a new prerelease immediately, brb

mookid8000:
Check this out: I just ran the `TestDispatchPerformance` test before and after your optimization.

Here's before:
```
.NET Core 2.1
	Avg s: 3,26354
	Avg msg/s: 30642

	Med s: 3,23851
	Med msg/s: 30878

.NET 4.6
	Avg s: 3,53083
	Avg msg/s: 28322

	Med s: 3,51766
	Med msg/s: 28428

.NET 4.5
	Avg s: 3,70674
	Avg msg/s: 26978

	Med s: 3,49214
	Med msg/s: 28636
```
As you can see, it's right past 30k msg/s on .NET Core and well below on full FX.

Here's AFTER your optimization:
```
.NET Core 2.1
	Avg s: 3,01399
	Avg msg/s: 33179

	Med s: 3,00548
	Med msg/s: 33273

.NET 4.6
	Avg s: 3,34372
	Avg msg/s: 29907

	Med s: 3,35941
	Med msg/s: 29767

.NET 4.5
	Avg s: 3,57061
	Avg msg/s: 28006

	Med s: 3,22244
	Med msg/s: 31032
```
This time it's well past 30k on .NET Core, and almost on 30k with full FX.

With this unscientific benchmark, I think it's safe to say that performance has improved 😎

mookid8000:
it's out in Rebus 6.0.0-b10 on NuGet.org in a few minutes

jods4:
Thanks for benchmarking!

I did this purely on a memory allocation perspective, it's nice to see it actually translates into a real perf improvement. 🎉
Even though it's only a tiny part of sending a message!

PR #825, opened 2019-08-26
Tiny optimization

jods4:
Get rid of 2 fields.
Including a ManualResetEvent, which wraps native kernel resources.

- `disposed` flag is redundant with the token source, which already indicates whether we stopped the async task or not.
- the costly kernel32 event is not required, as we can just wait on the task itself to complete! No need for synchronisation primitives here.

---


mookid8000:
cool, thanks 👍

mookid8000:
it's out now in Rebus 6.0.0-b12

Issue #883, opened 2020-06-03
Overloads for saga correlation methods


seankearon:
When we correlate messages that participate in sagas against their data, we use an Expression as the second parameter.  Like this example from the [wiki](https://github.com/rebus-org/Rebus/wiki/Coordinating-stuff-that-happens-over-time):

``` c#
    protected override void CorrelateMessages(ICorrelationConfig<CollectLegalInfoSagaData> config)
    {
        config.Correlate<CustomerCreated>(m => m.CustomerId, d => d.CrmCustomerId);
        config.Correlate<LegalInfoAcquiredInFirstSystem>(m => m.CorrId, d => d.CrmCustomerId);
        config.Correlate<LegalInfoAcquiredInSecondSystem>(m => m.CorrId, d => d.CrmCustomerId);
    }
```

I think this parameter is only used to identify the name of the property in the SagaData. 

Would it be sensible to have an overload for the second parameter that allowed us to pass a string?  For example:

``` c#
    protected override void CorrelateMessages(ICorrelationConfig<CollectLegalInfoSagaData> config)
    {
        config.Correlate<CustomerCreated>(m => m.CustomerId, nameof(CollectLegalInfoSagaData.CrmCustomerId));
    }
```
My motivation is to make it easier to use a saga from another language, F# in this case.

I think this would also apply to the other methods in [ICorrelationConfig.cs](https://github.com/rebus-org/Rebus/blob/master/Rebus/Sagas/ICorrelationConfig.cs) also.

mookid8000:
Yes, that would actually be great 🙂

seankearon:
Great!   How does [this](https://github.com/rebus-org/Rebus/pull/884) look?

mookid8000:
Fixed in Rebus 6.3.0 🤘

PR #884, opened 2020-06-04
Saga data property names can be specified as strings

seankearon:
Added overloads to allow saga data property names to be specified using strings.

#883

---


mookid8000:
Cool!

One thing though: Would it be possible to copy/paste some or all of the relevant tests and the relevant saga in `TestSagaCorrelation`, so that the new tests will automatically get executed by all implementors as soon as they update their dependency on the contracts NuGet package?

seankearon:
Sure - will look now...

seankearon:
I think I've done that 🤔 : 

I added a [parameter to the base class](https://github.com/seankearon/Rebus/blob/5ac3abb4b4146e6b4fa0c4b637c5a45446f72e5b/Rebus.Tests.Contracts/Sagas/TestSagaCorrelation.cs#L99) to allow the saga to switch from expressions to strings.  

Then I added a new [concrete fixture](https://github.com/seankearon/Rebus/blob/5ac3abb4b4146e6b4fa0c4b637c5a45446f72e5b/Rebus.Tests/Persistence/InMem/TestSagaCorrelationUsingSagaDataPropertyNamesInMem.cs#L7) with the parameter set so that is was called with the string option enabled.  

Should I be calling it differently?

mookid8000:
The problem is that all implementors must *remember* to add a new concrete test fixture to include the new tests.

E.g. Rebus.SqlServer currently has [`TestSagaCorrelationSql`](https://github.com/rebus-org/Rebus.SqlServer/blob/master/Rebus.SqlServer.Tests/Sagas/TestSagaCorrelationSql.cs) .... if it updates its reference to Rebus.Tests.Contracts, then it will not automatically get the new tests.... it will only do that if you add new tests to the `TestSagaCorrelation` class.

seankearon:
Aha! 💡 Thanks, Mogens.  I'll fix that asap.

seankearon:
The tests are now in the `TestSagaCorrelation`! 😺

mookid8000:
Excellent! It's out in Rebus 6.3.0 on NuGet.org now 🙂 

While it's actually strictly speaking a "breaking change" (because the `ICorrelationConfig<>` has had things added to it, and additional implementors are then made incomplete), I chose to only bump minor (6.2.1 👉 6.3.0), because I think it is very uncommon to implement `ICorrelationConfig`.

seankearon:
Fantastic - thanks, Mogens! 🍾

Issue #882, opened 2020-05-28
New Rebus Circuit-Breaker Feature


robpackwood:
This relates back to my StackOverflow question here: https://stackoverflow.com/questions/62069828/rebus-circuit-breaker-implementation

Basically, what I would like is configuration options on the RebusConfigurer class. Perhaps a ConfigureCircuitBreaker method with the following parameters:

- TrackingSeconds (int) - The period after which attempt/failure counts are reset.
- PercentFailureThreshold (int) - The percentage of attempts that must fail before the circuit breaker trips into an open state and reduces the worker count to 0.
- MinimumAttemptThreshold (int) - The number of attempts that must occur before the circuit breaker becomes active.
- ResetIntervalSeconds (int) - Sets a specific reset interval for the circuit to close after being tripped.

This would be so helpful for preventing thousands of messages going into the error queue unnecessarily.

mookid8000:
@seankearon yes! – I can definitely see how that would be useful

robpackwood:
I was the one who originally requested this feature. I have a duct-tape solution I am using locally for now. Hopefully, this isn't completely insane but what I did was is register a Lazy<IBus> to my IOC container and then basically wrapped the PoisonQueueErrorHandler in a class called CircuitBreakerErrorHandler and have some options in a local class I have called RebusConfig such as CircuitBreakerFailedMessageCount, CircuitBreakerTrackingIntervalSeconds, and CircuitBreakerResetIntervalSeconds.

`	public class CircuitBreakerErrorHandler : IErrorHandler, IInitializable
	{
		private readonly Lazy<IBus> _busLazy;
		private readonly ConcurrentQueue<DateTime> _errorDateTimes;
		private readonly PoisonQueueErrorHandler _poisonQueueErrorHandler;
		private readonly RebusConfig _rebusConfig;
		private int _busWorkerCount;
		private readonly object _lock = new object();

		public CircuitBreakerErrorHandler(
			Lazy<IBus> busLazy,
			RebusConfig rebusConfig,
			PoisonQueueErrorHandler poisonQueueErrorHandler)
		{
			_busLazy = busLazy;
			_rebusConfig = rebusConfig;
			_poisonQueueErrorHandler = poisonQueueErrorHandler;
			_errorDateTimes = new ConcurrentQueue<DateTime>();
		}

		public async Task HandlePoisonMessage(
			TransportMessage transportMessage, ITransactionContext transactionContext, Exception exception)
		{
[...]

PR #890, opened 2020-07-26
[Draft] Circuit breaker Feature

mclausen:
Hi Rebus Team,

The following PR is the first stab at implementing a circuit breaker feature described in #882. The feature is not entirely done, so I would love some help and feedback getting the last details in place.

My motivation for this is that, after the issue was mentioned we internally in our company had a couple of discussions on where this could be applied. We found some very interesting use-cases that could potentially save a ton of workload, clean-up and headaches. So without much consideration I went ahead and took the first stab at the feature and to see where i goes.

The idea is pretty simple, Based on a known exception Type, trip and trigger a circuit breaker. In a case where the circuit opens, we would `SetNumberOfWorkers(0)` to ensure that Rebus has stopped processing messages for the time being.

### Configuration
This is how i imagine how you would configure circuit breakers.
```cs
 var bus = Configure.With(...) 
     .Options(o => 
     { 
          o.SetCircuitBreakers(c => c.OpenOn<MyExternalServiceException>(attempts: 5
                                                                       , trackingPeriodInSeconds: 10
                                                                       , resetIntervalInSeconds: 300)); 
     })
     .Start(); 
```

#### ErrorHandler
The `IErrorHandler` is first triggered after a handler has failed 5 times, this means, that if we have a faulty service throwing an exception,  the service still getting taxed at least 5 times before we trip 1 attempt. 
If we configure circuit breaker to have 5 attempts within an interval, then the service will get taxed 25 times! 
**Question** is whether we should move this logic away from the `IErrorHandler` and into a `IIncomingStep`, to ensure that an external dependency  is taxed more than necessary?

#### Half-Open State
Circuit breaker patterns defines the concepts of half open. In case of Rebus i would believe that this transfers into changing the number of workers of Rebus. 

**Question** As for the ´MainCircuitBreaker´, it detects whether we are in a half open state. Is there a number of appropriate workers we want to set, or should this be a calculated value?

`ExceptionTypeCircuitBreaker` Responsibility is to keep track of the individual exceptions and when they occurred. For the time being it will only detect when to Open the circuit. In this case I feel a bit vague on what it means for the circuit breaker to be half open. I believe the Half open state should be set when the circuit has been Open, no errors occurred in the previous period, and no reset event has been set.
**Question** What does it mean for the `ExceptionTypeCircuitBreaker` to be Half-Open?

#### Resetting the circuit breakers
I would love to get some input on how we would do this? Does rebus have some background task support that can inspect the circuit breakers and see whether the need to be reset? I am all 👂 👂

Any and all feedback is of course welcome :raised_hands:

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1036,6 +1036,9 @@
 ## 6.2.1
 * Add extension method that makes it possible to manually dead-letter the message currently being handled
 
+## 6.3.0
+* Add ability to point out correlation properties of saga data by specifying them as strings - thanks [seankearon]
+
 ---
 
 [AndreaCuneo]: https://github.com/AndreaCuneo
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1039,6 +1039,9 @@
 ## 6.3.0
 * Add ability to point out correlation properties of saga data by specifying them as strings - thanks [seankearon]
 
+## 6.3.1
+* Fix bug that would prevent fail-fast exceptions to cause messages to fail fast when thrown in a 2nd level retry
+
 ---
 
 [AndreaCuneo]: https://github.com/AndreaCuneo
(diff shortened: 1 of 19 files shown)


mclausen:
I messed up in having by having outdated git author info, so i couldn't parse the CLA check. After trying to amend author name and rebase (which turned out to go horrible wrong), I've re-submitting a new  pr 👍

PR #891, opened 2020-08-02
[Draft] - Circuit Breaker feature

mclausen:
Hi Rebus Team,

The following PR is the first stab at implementing a circuit breaker feature described in #882. The feature is not entirely done, so I would love some help and feedback getting the last details in place.

My motivation for this is that, after the issue was mentioned we internally in our company had a couple of discussions on where this could be applied. We found some very interesting use-cases that could potentially save a ton of workload, clean-up and headaches. So without much consideration I went ahead and took the first stab at the feature and to see where i goes.

The idea is pretty simple, Based on a known exception Type, trip and trigger a circuit breaker. In a case where the circuit opens, we would `SetNumberOfWorkers(0)` to ensure that Rebus has stopped processing messages for the time being.

### Configuration
This is how i imagine how you would configure circuit breakers.
```cs
 var bus = Configure.With(...) 
     .Options(o => 
     { 
          o.EnableCircuitBreakers(c => c.OpenOn<MyExternalServiceException>(attempts: 5
                                                                       , trackingPeriodInSeconds: 10
                                                                       , resetIntervalInSeconds: 300)); 
     })
     .Start(); 
```

#### ErrorHandler
The `IErrorHandler` is first triggered after a handler has failed 5 times, this means, that if we have a faulty service throwing an exception,  the service still getting taxed at least 5 times before we trip 1 attempt. 
If we configure circuit breaker to have 5 attempts within an interval, then the service will get taxed 25 times! 
**Question** is whether we should move this logic away from the `IErrorHandler` and into a `IIncomingStep`, to ensure that an external dependency  is taxed more than necessary?

#### Half-Open State
Circuit breaker patterns defines the concepts of half open. In case of Rebus i would believe that this transfers into changing the number of workers of Rebus. 

**Question** As for the ´MainCircuitBreaker´, it detects whether we are in a half open state. Is there a number of appropriate workers we want to set, or should this be a calculated value?

`ExceptionTypeCircuitBreaker` Responsibility is to keep track of the individual exceptions and when they occurred. For the time being it will only detect when to Open the circuit. In this case I feel a bit vague on what it means for the circuit breaker to be half open. I believe the Half open state should be set when the circuit has been Open, no errors occurred in the previous period, and no reset event has been set.
**Question** What does it mean for the `ExceptionTypeCircuitBreaker` to be Half-Open?

#### Resetting the circuit breakers
I would love to get some input on how we would do this? Does rebus have some background task support that can inspect the circuit breakers and see whether the need to be reset? I am all 👂 👂

Any and all feedback is of course welcome :raised_hands:

---

### DIFF
--- a/Rebus.Tests/Rebus.Tests.csproj
+++ b/Rebus.Tests/Rebus.Tests.csproj
@@ -46,4 +46,7 @@
     <Reference Include="System.Core" />
     <Reference Include="Microsoft.CSharp" />
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="Retry\CircuitBreaker\" />
+  </ItemGroup>
 </Project>
--- a/Rebus.Tests/Retry/CircuitBreaker/CircuitBreakerTests.cs
+++ b/Rebus.Tests/Retry/CircuitBreaker/CircuitBreakerTests.cs
@@ -1,0 +1,51 @@
+﻿using NUnit.Framework;
+using Rebus.Activation;
+using Rebus.Config;
+using Rebus.Retry.CircuitBreaker;
+using Rebus.Tests.Contracts;
+using Rebus.Transport.InMem;
+using System;
+using System.Threading.Tasks;
+
+namespace Rebus.Tests.Retry.CircuitBreaker
+{
+    [TestFixture]
+    public class CircuitBreakerTests : FixtureBase
+    {
+        [Test]
+        [Ignore("Temporary ignored, For some reason this test fails, even though worker count explicity set to 0, but worker count returns 1")]
+        public async Task Test()
+        {
+            var network = new InMemNetwork();
+
+            var receiver = Using(new BuiltinHandlerActivator());
+
+            var bus = Configure.With(receiver)
+                  .Transport(t => t.UseInMemoryTransport(network, "queue-a"))
+                  .Options(o =>
+                      {
+                          o.SetCircuitBreakers(c => c.OpenOn<MyCustomException>(1, trackingPeriodInSeconds: 10));
+                      }
+                  )
+                  .Start();
+
+            receiver.Handle<string>(async (buss, context, message) =>
+            {
+                await Task.FromResult(0);
+                throw new MyCustomException();
+            });
+
+            await bus.SendLocal("Uh oh, This is not gonna go well!");
+
+            await Task.Delay(5000);
+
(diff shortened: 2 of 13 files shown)


mookid8000:
Hi @mclausen , sorry for being so slow to respond to this.... my only excuse is that I've been busy 😎 doing 🍺 other 🏖️ things.

First off, I want to say I'm impressed with the quality of this work – not often do I get PRs where I basically don't feel like changing a single dot. 😸 

I have one wish though: I would like this to become a separate package, e.g. something like Rebus.CircuitBreaker, because it so neatly doesn't change a single thing in Rebus core, and so I think it belongs in its own project, which can be independently developed/fixed etc.

And one more thing: You ask whether `IErrorHandler` is the best place to hook the circuit break functionality in... actually, I think there's a better place: `IErrorTracker`! 😁 It might be a little bit confusing, but the responsibility is divided like this:

* `IErrorTracker` tracks delivery attempts (i.e. which exceptions occurred when for a given message ID), and
* `IErrorHandler` is invoked once the tracker passes judgement that the message has failed too many times

This way, I believe a decorator of the `RegisterError` method of [`IErrorTracker`](https://github.com/rebus-org/Rebus/blob/master/Rebus/Retry/IErrorTracker.cs) would be the optimal place to trip the circuit breaker.

So.... what do you say I go and set up a separate repo for Rebus.CircuitBreaker?

mookid8000:
Ok I've already prepared a repo for it here 👉  https://github.com/rebus-org/Rebus.CircuitBreaker

mclausen:
Thank you for the kind words @mookid8000, but you know I am learning from the best 🙌 

Sure thing, ill move it at first opportunity 👍 

In regards to the `BusLifetimeEvents.CircuitBreakerChangedState` and `CircuitBreakerState` are these still needed? If so I believe we need to have them added to Rebus core. Alternatively we can create a `CircuitBreakerEvents` for exposing events from within the Circuit Breaker.  In reality this is not really needed, though it would be nice for the consumer of the package to specify an action if the circuit get tripped 🤔.

mookid8000:
Cool cool!

And yes, I believe adding a `CircuitBreakerEvents` thing to Injectionist is the right way to go.

mclausen:
Discusstion continued in https://github.com/rebus-org/Rebus.CircuitBreaker/pull/1

PR #894, opened 2020-08-29
IAsyncTaskFactory: use TimeSpan for interval

AndreaCuneo:
Replace integer seconds parameter with `TimeSpan` avoiding unnecessary conversion between `TimeSpan` and `int` in `HandleDeferredMessagesStep`.

It also removes a strange behaviour in case of `DueTimeoutsPollInterval` is set with <0.5 `TotalSeconds`, that would cause to call `IAsyncTaskFactory` with intervalSeconds:0, thus clamped to `100ms` by Interval setter on `TplAsyncFactory`.

Also fix `SystemThreadingTimerAsyncTaskFactory` ignoring interval parameter.

---


mookid8000:
cool, thanks 👍

PR #897, opened 2020-09-10
added saga serializer configuration to be able to configure saga seri…

mathiasnohall:
We need to be able to configure how the data is stored in the rebus saga

### DIFF
--- a/Rebus/Sagas/ISagaSerializer.cs
+++ b/Rebus/Sagas/ISagaSerializer.cs
@@ -1,0 +1,28 @@
+﻿namespace Rebus.Sagas
+{
+    /// <summary>
+    /// Serializer used to serialize and deserialize saga data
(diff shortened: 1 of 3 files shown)


mookid8000:
hi @mathiasnohall , thanks for picking this one up!

However, I'm not sure I agree that the saga serializer should be in the Rebus (core) project.

As you can see, it's not even used anywhere, besides in the configuration methods.

The reason I've been hesitant to add a saga serializer mechanism in the past, is because it only makes sense for saga persisters that need serialization, but e.g. with Rebus.MongoDb the Mongo driver's BSON serializer will make it feel like you're saving objects to the database, and so it doesn't need any kind of upfront serialization – you effectively get to save/update objects.

SO.... could you maybe move the saga serializer to Rebus.SqlServer?

PR #927, opened 2021-01-26
Add support for async encryption

### LINKED ISSUE #926: Async IEncryptor
Hi,

It's common for our applications to have access to an external key store, such as Azure Keyvault. As such we'd like to utilize that for message encryption. Ideally we'd create a custom implementation of [`IEncryptor`](https://github.com/rebus-org/Rebus/blob/master/Rebus/Encryption/IEncryptor.cs), but its method declarations do not allow the asynchronous implementations we'd need. 

Do you have any thoughts regarding this?

I'm thinking it could be resolved in a non-breaking way by introducing a new interface `IAsyncEncryptor` and register a default implementation that wraps `IEncryptor`, though I haven't fully examined the code. 

I'd be happy to take a shot at a PR if interested.

asleire:
Closes #926

---

### DIFF
--- a/Rebus.Tests/Encryption/TestCustomAsyncEncryption.cs
+++ b/Rebus.Tests/Encryption/TestCustomAsyncEncryption.cs
@@ -2,6 +2,7 @@
 using System.Collections.Concurrent;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
 using NUnit.Framework;
 using Rebus.Activation;
 using Rebus.Config;
@@ -9,7 +10,6 @@ using Rebus.Encryption;
 using Rebus.Messages;
 using Rebus.Tests.Contracts;
 using Rebus.Tests.Contracts.Extensions;
-using Rebus.Tests.Extensions;
 using Rebus.Tests.Transport;
 using Rebus.Transport;
 using Rebus.Transport.InMem;
@@ -18,7 +18,7 @@ using Rebus.Transport.InMem;
 namespace Rebus.Tests.Encryption
 {
     [TestFixture]
-    public class TestCustomEncryption : FixtureBase
+    public class TestCustomAsyncEncryption : FixtureBase
     {
         TransportTap _transportTap;
         BuiltinHandlerActivator _activator;
@@ -34,7 +34,7 @@ namespace Rebus.Tests.Encryption
                 .Transport(t => t.UseInMemoryTransport(new InMemNetwork(), "custom-encryption"))
                 .Options(o =>
                 {
-                    o.EnableCustomEncryption()
+                    o.EnableCustomAsyncEncryption()
                         .Register(c => new SillyEncryptor());
 
                     o.Decorate<ITransport>(c =>
@@ -59,14 +59,14 @@ namespace Rebus.Tests.Encryption
 
             _activator.Handle<string>(async message =>
             {
-                if (message != "hej") throw new ArgumentException("not the right message!!");
+                if (message != "hei") throw new ArgumentException("not the right message!!");
 
                 gotMessage.Set();
             });
-            
+
             _starter.Start();
 
-            _activator.Bus.SendLocal("hej").Wait();
+            _activator.Bus.SendLocal("hei").Wait();
 
             gotMessage.WaitOrDie(TimeSpan.FromSeconds(3));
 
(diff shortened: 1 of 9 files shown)


Rebus/Encryption/DecryptMessagesIncomingStep.cs
```diff
@@ -15,18 +15,18 @@ namespace Rebus.Encryption
...
     public class DecryptMessagesIncomingStep : IIncomingStep
     {
-        readonly IEncryptor _encryptor;
+        readonly IAsyncEncryptor _encryptor;
 
         /// <summary>
         /// Constructs the step with the given encryptor
         /// </summary>
-        public DecryptMessagesIncomingStep(IEncryptor encryptor)
+        public DecryptMessagesIncomingStep(IAsyncEncryptor encryptor)
```
asleire, line 23: I'm somewhat concerned about changing the constructor since this is a public class. It could cause issues if someone is using the `IEncryptor` constructor in their code.

An alternative is keeping the old constructor so that the step has two constructors. I'm not a fan of multiple service constructors, but at least it would be safe

Rebus/Encryption/EncryptMessagesOutgoingStep.cs
```diff
@@ -37,10 +37,11 @@ public async Task Process(OutgoingStepContext context, Func<Task> next)
 
             var headers = transportMessage.Headers.Clone();
             var bodyBytes = transportMessage.Body;
-            var encryptedData = _encryptor.Encrypt(bodyBytes);
+            var encryptedData = await _encryptor.Encrypt(bodyBytes);
 
             headers[EncryptionHeaders.ContentEncryption] = _encryptor.ContentEncryptionValue;
             headers[EncryptionHeaders.ContentInitializationVector] = Convert.ToBase64String(encryptedData.Iv);
+            headers[EncryptionHeaders.KeyId] = encryptedData.KeyId;
```
asleire, line 44: I didn't dig deep enough to see - will empty headers be omitted? If not, I'd like to add a null check here

Rebus.Tests/Encryption/TestCustomAsyncEncryption.cs
```diff
@@ -0,0 +1,106 @@
...
+            var transportMessages = new ConcurrentQueue<TransportMessage>();
+
+            _transportTap.MessageReceived += transportMessages.Enqueue;
+            _transportTap.MessageSent += transportMessages.Enqueue;
+
+            var gotMessage = new ManualResetEvent(false);
+
+            _activator.Handle<string>(async message =>
+            {
+                if (message != "hej") throw new ArgumentException("not the right message!!");
```
mookid8000, line 62: Shouldn't that be "hei"? 😁

mookid8000 (approved):

Rebus/Encryption/DecryptMessagesIncomingStep.cs
mookid8000: Yeah.... well.... while I *lean* towards SemVer compliance, I'm too much of a pragmatic to care about this. 

I would say that it is extremely uncommon for anyone to have taken a compile-time dependency on the signature of this particular constructor.

Let's just push this as it is, and then we'll deal with it later if it causes anyone trouble.

Rebus/Encryption/EncryptMessagesOutgoingStep.cs
mookid8000: >  will empty headers be omitted?

No they won't. So yeah, it would be nice to skip setting the header if the value is `null`.

asleire: [Fix'd](https://github.com/rebus-org/Rebus/pull/927/commits/a2377065724fdc7178b3a4629cbbab6030cdb655#diff-cea16257463d1bb4ad07afbb6e6ec096491aa017766dba3642ac2bb69d8ebae0R47)

Rebus/Encryption/EncryptMessagesOutgoingStep.cs
```diff
@@ -37,10 +38,13 @@ public async Task Process(OutgoingStepContext context, Func<Task> next)
...
+
+            headers = headers.Where(kvp => !string.IsNullOrEmpty(kvp.Value)).ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
```
mookid8000, line 47: I think you should just skip setting the `EncryptionHeaders.KeyId` header if `encryptedData.KeyId` is `null`. Doing this has the potential of changing other headers, which would be quite unexpected 😁

mookid8000 (changes requested):
Could you maybe leave the other headers alone? 😆

Rebus/Encryption/EncryptMessagesOutgoingStep.cs
asleire: Fix'd!

mookid8000 (approved):

PR #942, opened 2021-02-19
Add support for mapping all types under a specific namespace in an assembly

kendallb:
Make sure to only map classes for message types when doing assembly mapping, and added support for mapping all types under a namespace within an assembly.

---


mookid8000:
Excellent! Will be out in Rebus 6.5.0 some time today 🙂

PR #948, opened 2021-02-25
Add support for bailing from a failed message handler 

kendallb:
In order to allow a failed message handler to abort and punt the message to the poison queue, if we throw a FailFastException that also ends up in the exception log for the message in the poison queue, which is annoying. This patch will no log any FailFastException's in second level retry handlers to avoid that. Used like this:

```C#
    var deferCount = CM.ToInt(message.Headers.GetValueOrDefault(Headers.DeferCount));
    if (deferCount >= 3) {
        throw new FailFastException("Too many errors!");
    }
    await Bus.Advanced.TransportMessage.Defer(TimeSpan.FromSeconds(5));
```

---

### DIFF
--- a/Rebus/Retry/Simple/SimpleRetryStrategyStep.cs
+++ b/Rebus/Retry/Simple/SimpleRetryStrategyStep.cs
@@ -138,7 +138,10 @@ If the maximum number of delivery attempts is reached, the message is moved to t
             }
             catch (Exception exception)
             {
-                _errorTracker.RegisterError(identifierToTrackMessageBy, exception);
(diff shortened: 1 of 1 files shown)


mookid8000:
I think this messes with functionality as it is, so it's not so straightforward to make this change.

Also, did you consider simply dead-lettering the message manually?

It can be done like this:
```csharp
if (deferCount >= 3) {
	await Bus.Advanced.TransportMessage.DeadLetter(errorDetails: $"Couldn't do it even after {deferCount} deferrals");
}
```

kendallb:
Ahh, I will try that! Can you update the documentation about second level retries fo include that?

PR #950, opened 2021-03-05
Use a GUID to avoid crashes during startup for Filesystem Databus

kendallb:
Use a GUID to generate the test file for checking the directory is writeable. If you have multiple threads all starting up within the same second and accessing the same shared directory location, it can cause crashes. In our case we have three web apps that all start up at once. One for the front end web site (one way bus), one for the back end web site (full bus with handlers) and one for our POS front end site (one way bus). They can all spin up at the same time, and hence crash out as the filename used in all three threads can end up being the same due to the timestamp system used. Changing to a GUID will mean they will never conflict.

---

### DIFF
--- a/Rebus/DataBus/FileSystem/FileSystemDataBusStorage.cs
+++ b/Rebus/DataBus/FileSystem/FileSystemDataBusStorage.cs
@@ -223,7 +223,7 @@ namespace Rebus.DataBus.FileSystem
             }
             catch (IOException)
             {
-                // this exception is most likely caused by a locked file because someone else is updating the 
+                // this exception is most likely caused by a locked file because someone else is updating the
                 // last read time  - that's ok :)
             }
             catch (Exception exception)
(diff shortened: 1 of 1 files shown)


kendallb:
BTW, the other alternative is to change this to do it the way Saga storage does it, in which case the writeable access check is only done if the directory is not found, rather than on every startup? You can assume if the directory is there, it's probably good to go?

mookid8000:
Using a GUID is a great way to solve this problem. 🙂 

> BTW, the other alternative is to change this to do it the way Saga storage does it, in which case the writeable access check is only done if the directory is not found, rather than on every startup? You can assume if the directory is there, it's probably good to go?

This would not catch the case where the currently executing process does not have read/write access to the directory.

mookid8000:
It's out in Rebus 6.5.4

kendallb:
> This would not catch the case where the currently executing process does not have read/write access to the directory.

Yeah I thought about that also, and made me wonder if the saga storage should probably be doing the checks the way the data bus does for the same reason?

mookid8000:
It probably should 🙂

kendallb:
Created a new PR for you.

https://github.com/rebus-org/Rebus/pull/952

PR #952, opened 2021-03-07
Centralized the filesystem writeability checks

kendallb:
Centralized the filesystem writeability checks to use the same code for both databus and saga snapshots. Did not change anything related to other filesystem code like transport, since it was not currently doing that.

---


mookid8000:
Cool 🙂 it's out in Rebus 6.5.5

PR #960, opened 2021-03-17
Need a way for the exclusive saga lock instance to be pulled from injectionist

kendallb:
Need a way for the exclusive saga lock instance to be pulled from injectionist so we can register it and have it's lifecycle managed.

---


kendallb:
Not that this will not build until Rebus has been updated with the missing dependencies due to interface changes.

kendallb:
Hmm, says Rebus.Tests.Extensions.ManualDeadlettering.CannotDeadletterTwice fails, but when I run it myself, it works? Perhaps that test is a bit flaky?

kendallb:
Working on another and better way to do this. It's more changes in Rebus, but I think it's cleaner ...

PR #962, opened 2021-03-19
Much better implementation of exclusive locks interface and configuration

kendallb:
Added support for registered exclusive locking interface with the configuration and build a new configuration section to support it.

Updated the step handler to look for the registered handler and use that if found, otherwise fall back on the default semaphore slim model.

Added support for jittered wait delays for lockers that cannot do an async wait when aquiring the lock (ConurrentDictionary and MySQL locks).

Updated all unit tests and made sure the modify the customer locker to work correctly. Rebased unit tests to have a common base to test both the semaphore slim code and the concurrent dictionary code which effectively tests the API's to be used by MySQL.

Added function I need for MySQL locks to check if a lock is aquired without taking the lock or blocking.

---


kendallb:
I am tempted to add a 'payload' to the lock table, as then we could use it for implementing async reply functionality across web server nodes? I am tempted to go that way to get rid of our REST calls between machines, as while it might introduce a bit more latency, I think in the long run it will make it easier to pick up the bits when things go wrong (and when stuff is back, if we have messages persisted then much of it will end up running to completion, leaving the system in a much better state than we get now when REST calls fail out as they can never be replayed).

Thoughts?

mookid8000:
Hi @kendallb , sorry I am not sure that this is the right approach... first off, I'd like to talk about why you need the locking API to be made publicly accessible?

kendallb:
I wanted it publicly accessible because it’s extremely useful for anyone using Rebus to be able to use the locks. Since it’s registered internally in Rebus I was not able to find a way for me to access it for external use via Autofac any other way. 

The alternative if we don’t want to make it publicly accessible I will need to clone it into my own code and create my own instance of it (and use a different table). That’s certainly plausible but it seems so useful. I can’t just create an instance of the MySQL locker implementation either because it has internal classes used inside it like the clean task helper and logger and there is no support I can find for getting access to a class instance from the injectionist from outside code.

mookid8000:
> I wanted it publicly accessible because it’s extremely useful for anyone using Rebus to be able to use the locks. Since it’s registered internally in Rebus I was not able to find a way for me to access it for external use via Autofac any other way.

I am not sure I can imagine any use cases where I would compete for the same locks as Rebus would – remember, the locks are based on a weird, technical naming scheme consisting of a string generated from an anonymous object that points to a saga data property name and the corresponding value from the incoming message.

If you do want to have locks available to your own code, there's nothing that prevents you from using your own locking mechanism, e.g. `IAsyncLockingThing`, which you can then have Rebus use
```csharp
services.AddSingleton<IAsyncLockingThing, YourAsyncLockingThing>();

services.AddRebus(
    (configure, provider) => configure
        .Transport(...)
        .Options(o => o.UseThisLockThing(provider.GetRequiredService<IAsyncLockingThing>()))
);
```
and then also use in your handlers and other application services by having it injected via `IAsyncLockingThing`.

You would of course need to implement `UseThisLockThing` as an extension that registers an adapter from Rebus' `IExclusiveSagaAccessLock` to `IAsyncLockingThing`, but that would probably not be too hard.

How does that sound?

As you can probably see, I am very reluctant to extend APIs in Rebus, because they are a burden to maintain for all eternity (or at least for a very long time... 😅) – so I will go to great lengths to try alternative solutions before doing so... 🙂

kendallb:
Sure I thought about doing that also as the interface to for the saga locking allows you to pass in your own instance, but then I cannot use all the nice stuff Rebus has that I can use for the MySQL implementation like logging and the background cleanup thread. I would need my own implementation for that. As I mentioned above, I did not see a way to be able to build my own instance of the MySQL locker class and pass in those dependencies as they are all internal to Rebus.

Or the alternative which I tried to figure out, of somehow being able to pull out the instance of the locker to register it with Autofac so I can inject it into code that needs it.

But you are right, and I don't really like making it a public API either. Felt messy to me. So we should come up with another solution.

The simplest solution is to make it all internal again and I will just clone the locker and build my own implementation for our own messaging purposes. I did make it so the locker can use a prefix for the lock names in the table (which are now just ID's in the hash space as suggested, not the full lock name). And maybe that's better anyway so that only locks used internally by Rebus will be stored in there.

The other alternative to this which would eliminate my need to use the locker class at all in our code is to add in the payload to the MySQL lock table, which can be used to store the reply message for Rebus.Async. That way I can get Rebus.Async to work with cross process semaphore locks, and not have to use the concurrent dictionary implementation that will only work for in process support (not across web nodes for instance).

What I am using the locks for at the moment is the ability to submit a 'sync' message on the message bus with the same ordering key for critical work, and then be able to wait until the messages for that order have all been processed. That's turned out to be critical for certain UI pages to work, if they submit background operations that need to run in another context. In our case submitting messages to our primary web site database, which then needs to make order updates and post messages back to push those changes back into the distribution center database. If the UI page completes and displays the result of the operation before the messages have been processed, the user sees a page that looks like nothing was applied. 

At the moment this is done using the locking table and works nicely, but if I extended Rebus.Async to use the MySQL locker then I could simply make that sync message an inline message/reply with a timeout. And then I can also use it to replace REST calls. In both cases (REST or Message bus sync messages), if it times out, we throw a special exception so we know something timed out and can handle it in higher level code. But the biggest reason I want to use Rebus.Async to potentially replace our REST calls with inline message/reply handling is that when stuff does go bad, we have a good idea of where it went wrong and a better handler on how to 'pick up' where we left off. With REST, while it has retries if it does time out, all we have left over is the crash logs and we don't know how far it got. With message processing the messages are actually processed out of process from the web processes where the REST handlers reside, which also means they do not get mangled if the web processes are overloaded for some reason (DDoS attack etc).

I will work on fleshing out the addition of the payload and make the API's private again and let you take another look when I am done.

mookid8000:
> Sure I thought about doing that also as the interface to for the saga locking allows you to pass in your own instance, but then I cannot use all the nice stuff Rebus has that I can use for the MySQL implementation like logging and the background cleanup thread. I would need my own implementation for that. As I mentioned above, I did not see a way to be able to build my own instance of the MySQL locker class and pass in those dependencies as they are all internal to Rebus.
> 
> Or the alternative which I tried to figure out, of somehow being able to pull out the instance of the locker to register it with Autofac so I can inject it into code that needs it.
> 
> But you are right, and I don't really like making it a public API either. Felt messy to me. So we should come up with another solution.
> 
> The simplest solution is to make it all internal again and I will just clone the locker and build my own implementation for our own messaging purposes. I did make it so the locker can use a prefix for the lock names in the table (which are now just ID's in the hash space as suggested, not the full lock name). And maybe that's better anyway so that only locks used internally by Rebus will be stored in there.

That sounds like a better solution to me.

> The other alternative to this which would eliminate my need to use the locker class at all in our code is to add in the payload to the MySQL lock table, which can be used to store the reply message for Rebus.Async. That way I can get Rebus.Async to work with cross process semaphore locks, and not have to use the concurrent dictionary implementation that will only work for in process support (not across web nodes for instance).
> 

I don't understand a thing of what you're saying here. 🤔 I mean, I understand that Rebus.Async does not work, if you're on a web farm, because the reply needs to go back to a specific instance – but OTOH that might call for a uniquely named input queue (i.e. one queue per web instance) instead of bringing in the lock table.

> What I am using the locks for at the moment is the ability to submit a 'sync' message on the message bus with the same ordering key for critical work, and then be able to wait until the messages for that order have all been processed. That's turned out to be critical for certain UI pages to work, if they submit background operations that need to run in another context. In our case submitting messages to our primary web site database, which then needs to make order updates and post messages back to push those changes back into the distribution center database. If the UI page completes and displays the result of the operation before the messages have been processed, the user sees a page that looks like nothing was applied.
> 
> At the moment this is done using the locking table and works nicely, but if I extended Rebus.Async to use the MySQL locker then I could simply make that sync message an inline message/reply with a timeout. And then I can also use it to replace REST calls. In both cases (REST or Message bus sync messages), if it times out, we throw a special exception so we know something timed out and can handle it in higher level code. But the biggest reason I want to use Rebus.Async to potentially replace our REST calls with inline message/reply handling is that when stuff does go bad, we have a good idea of where it went wrong and a better handler on how to 'pick up' where we left off. With REST, while it has retries if it does time out, all we have left over is the crash logs and we don't know how far it got. With message processing the messages are actually processed out of process from the web processes where the REST handlers reside, which also means they do not get mangled if the web processes are overloaded for some reason (DDoS attack etc).
>

It does sound a little bit like a fairly complicated device. If I were you, I would take some [hammock time](https://melreams.com/2017/05/rich-hickey-hammock-driven-development/) and see if I could come up with something simpler.

kendallb:
Yeah it is a bit complicated but it works well. If the traffic eventually grows we will need to let stuff be eventually consistent from the web UI point of view, but our users are pretty impatient so they want to know it actually did the work :) So for now a sync message works quite nicely and is much more resilient than REST calls that can fail.

mookid8000:
Hi @kendallb , I'm sorry but I am still not convinced that the lock should be added to the advanced API.

It's too big a change, too much to add, to solve too narrow a problem.

I know this is a bummer to you. Let's talk about how we solve the problem you're struggling with, with an alternative design.

kendallb:
Yeah no worries. We are using it at the moment and I do plan to re-write it but have not gotten around to it. I hope to get back to that soon.

PR #969, opened 2021-04-29
Release/7

renemadsen:
- Updates to make it work with net 5.0
- Added github actions.
- Added dependabot actions as recommended by github.

You should be able to select squash all commits in this PR, by doing https://github.blog/2016-04-01-squash-your-commits/

---


mookid8000:
Hi @renemadsen , sorry for taking so long to process your PR. I think we had this discussion in another Rebus repository, but I feel I have to bring it up again.

First off, this is a BIG pull request to process. I love that you want to contribute, but it makes it very hard for me when multiple things (porting to .NET 5, adding GitHub actions, updating lots of packages, etc.) are bundled into one single PR. 

Next, it's not possible for me to decide which of the part I'd like to accept, because it's just one big PR that can be accepted/rejected as a whole.

Last, if you'd like to contribute, we should talk about it before you do too much work. It's great if you maintain your own fork of these things, but it takes a little bit more consideration when adding things back into the mainlain.

So with that, I'm closing this PR, in the hope that you will bring up what you think needs to be done in an issue. Thanks for your time 🙂

PR #971, opened 2021-05-10
Added support for distributed exclusive saga locks via MySQL tables (part deux)

kendallb:
Changed the implementation to keep the locking all private to Rebus. If this is acceptable I will push the necessary changes to the MySQL branch tomorrow also.

---


mookid8000:
Sorry! I accepted this PR by accident, so I had to revert it.

I don't think Rebus should be concerned with locking at the top level – if locks are required (e.g. around sagas), they're an implementation detail for the saga configuration.

Therefore, `RebusConfigurer` should not have a `.Locking(...)` configuration API.

Sorry for the confusion.

kendallb:
Right, I think I had it setup originally with the locking on the sagas only, but since it seemed useful to be able use the locking for things other than just saga locks, so I left it that way. But I can certainly move the API to configure it to just sagas. I have my own version of the class we use in our code now rather than sharing this one.

kendallb:
Just fixed some bugs in the MySQL transport. I need to get back to this so we can switch back to the official builds again. I will try to take a look at making that stuff internal again shortly. Just need to steal the bits I need for our own code, and did most of that already.

kendallb:
I am reverting this now as requested, but the way I originally set it up, IExclusiveAccessLock can be used for anything that needs exclusive access locks, not just for Sagas. Hence the reason I made it a top level API. Currently it is only used for Sagas sure, but it's quite useful in many scenarios to be able to have a distributed, exclusive lock system. We are not using it internally anymore (we re-implemented the locker in our own code separately for our own use case outside of Sagas), but it still feels useful to me to have this at the top level.

Either way I will push a new PR for the new code with this reverted (for both Rebus and MySQL).

kendallb:
Ugh, this is much harder to do without a way to register a locker somewhere in the configurer, so that it's lifecycle can be handled and it can be created as necessary when needed. Otherwise the call to enable SagaLocks needs you to pass in a locker, which means it's not so easy to build the configuration needed to materialize it. It would be a lot easier if we can make the IExclusiveAccesssLock API part of the top level configurer.

kendallb:
This is what I mean:

https://github.com/kendallb/Rebus.MySql/commit/e6a467045479c04ec98b01992f82040950118d23

To configure the saga locking in MySQL due to all the need for table management etc, with the locking API it was a matter of registering the MySQL locker, and then telling the Saga configuration to use it. Without the locking API, we now need to know a whole lot more about how to materialize the MySqlExclusiveAccessLock class within the user code, which while it works, it feels messy to me:

```C#
.Locking(l => l.UseMySql(new MySqlExclusiveAccessLockOptions(MySqlTestHelper.ConnectionString), _lockTableName))
.Sagas(s =>
{
    s.StoreInMySql(MySqlTestHelper.ConnectionString, _dataTableName, _indexTableName);
    s.EnforceExclusiveAccessViaLocker();
```

feels a whole lot cleaner to me than this:

```C#
.Sagas(s =>
{
    s.StoreInMySql(MySqlTestHelper.ConnectionString, _dataTableName, _indexTableName);
    s.EnforceExclusiveAccess(c =>
    {
        var options = new MySqlExclusiveAccessLockOptions(MySqlTestHelper.ConnectionString);
        var rebusLoggerFactory = c.Get<IRebusLoggerFactory>();
        var connectionProvider = options.ConnectionProviderFactory(c);
        var asyncTaskFactory = c.Get<IAsyncTaskFactory>();
        var rebusTime = c.Get<IRebusTime>();
        var locker = new MySqlExclusiveAccessLock(connectionProvider, _lockTableName, rebusLoggerFactory, asyncTaskFactory, rebusTime, options);
        if (options.EnsureTablesAreCreated)
        {
            locker.EnsureTableIsCreated();
        }
        return locker;
    }, _lockTableName);
```

kendallb:
Here is the new PR

PR #1000, opened 2021-11-23
Added support for distributed exclusive saga locks via MySQL tables

kendallb:
This version gets rid of the public .Locking configuration API.

---


mookid8000:
Cool! 🙂 

Since this is a breaking change it'll be  available in Rebus 7.

I've released it as Rebus 7.0.0-b01, which is available on NuGet.org now

kendallb:
Great, thanks! I can update the MySQL tree to build against it now. What would be the ETA for the 7.0.0 release?

mookid8000:
I would like to do a couple of iterations over Rebus' code before releasing a final version 7, so my guess is that it will be out in a couple of weeks

kendallb:
Sounds good to me.

Issue #1014, opened 2022-02-08
hypothesist for integration tests


riezebosch:
After developing the 3rd framework-specific implementation of a blocking test handler, I managed to abstract that into a dedicated library. On top of that, I added framework-specific adapters to make the future assertions lean, self-describing, and with nice error messages.

```csharp
var hypothesis = Hypothesis
    .For<TMessage>()
    .Any(m => m == ...);

using var activator = new BuiltinHandlerActivator()
    .Register(hypothesis.AsHandler);
```

```csharp
await hypothesis
    .Validate(TimeSpan.FromSeconds(2));
```

This would replace the `ManualResetEvent` logic described in the [wiki](https://github.com/rebus-org/Rebus/wiki/How-to-create-integration-tests).

Check out the [hypothesist library](https://github.com/riezebosch/hypothesist/) and the [Rebus adapter](https://github.com/riezebosch/hypothesist/tree/main/Rebus).

I'd be happy to incorporate it into the wiki and use it in an upcoming PR for a scenario I would like to suggest.

mookid8000:
While this no doubt looks pretty neat, I am afraid it could add a layer of confusion to newcomers as to where the boundaries are between Rebus and not-Rebus.

It's a clear goal for Rebus to be welcoming to new users, and so I think it's very important that the code demonstrated relies mostly on Rebus combined with pure .NET BCL, and much less important that it's short.

If you like, you can add a mention and a link to Hypothesist to the wiki page above, but I'd like the example with `ManualResetEvent` to stay intact. 

I hope this makes sense to you. 🙂 

Sorry for taking so long to process this, btw.

PR #1015, opened 2022-02-08
feat(one-way): Configure.OneWayClient() with EmptyActivator

### LINKED ISSUE #1013: activator required for one way clients
Would it make sense to include an activator stub for one-way clients since you won't be activating anything?

```csharp
public class EmptyActivator : IHandlerActivator
{
    Task<IEnumerable<IHandleMessages<TMessage>>> IHandlerActivator.GetHandlers<TMessage>(TMessage message, ITransactionContext transactionContext) => 
        Task.FromResult(Enumerable.Empty<IHandleMessages<TMessage>>());
}
```

So you can do without disposing:
```csharp
Configure.With(new EmptyActivator())
    .Transport(t => t.UseRabbitMqAsOneWayClient(_container.ConnectionString))
    .Start();
```

riezebosch:
I slipt [Hypothesist](https://github.com/riezebosch/hypothesist/) in because of the easy integration testing, see #1014. For sure I'm open to rewriting it into `ManualResetEvent` or `TaskCompletionSource` if you'd prefer that.

I'm aware of the probably differentiating setup of the test but wasn't too sure how to fit it into the designated setup.

Fixes #1013.

---

### DIFF
--- a/Rebus.Tests/Activation/OneWay.cs
+++ b/Rebus.Tests/Activation/OneWay.cs
@@ -1,0 +1,61 @@
+using System;
+using System.Threading.Tasks;
+using Hypothesist;
+using Hypothesist.Rebus;
+using NUnit.Framework;
+using Rebus.Activation;
+using Rebus.Bus;
+using Rebus.Config;
+using Rebus.Persistence.InMem;
+using Rebus.Routing.TypeBased;
+using Rebus.Tests.Contracts;
+using Rebus.Transport.InMem;
+
+namespace Rebus.Tests.Activation;
+
+[TestFixture]
+public class ClientOnly : FixtureBase
+{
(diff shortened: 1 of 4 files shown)


mookid8000:
Great 🙂 also, I didn't know about Hypothesist – looks pretty neat!

I've accepted your PR and made a few adjustments, which I hope you're OK with:

* I made `EmptyActivator` internal, because `Configure.OneWayClient()` makes for a perfect entry point now
* I added some validation to ensure that the configured transport is in fact a one-way transport
* I fixed the test – you were close, you just needed to let the publisher and the subscriber share an in-mem subscription store 🙂 

It's out as Rebus 7.0.0-b05 on NuGet.org now.

Thanks for contributing!

PR #1018, opened 2022-02-15
Use a dedicated transaction context for express messages

nativenolde:
With express messaging, we can trade reliability for performance to deliver those messages as quickly as possible.
Therefore we should use a dedicated one-time single-purpose transaction context instead of the existing one.

Consider a message-handling scenario that takes a couple of seconds and needs to publish some events along the work, whereby being notified as soon as possible is more important than having top-notch reliability.

---

### DIFF
--- a/Rebus/Bus/MessageExtensions.cs
+++ b/Rebus/Bus/MessageExtensions.cs
@@ -129,6 +129,26 @@ public static class MessageExtensions
     }
 
     /// <summary>
+    /// Determines whether to trade reliability for performance to deliver this message as quickly as possible
+    /// </summary>
+    /// <param name="message"></param>
+    /// <returns></returns>
+    public static bool IsExpressMessage(this Message message)
+    {
+        return IsExpressMessage(message.Headers);
+    }
+
+    /// <summary>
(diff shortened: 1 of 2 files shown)


mookid8000:
I am not sure I can accept this change because of the change in semantics: It would make outgoing messages with the express header "escape" the current Rebus transaction context, which is either a Rebus handler or a manually created `RebusTransactionScope`.

Moreover, I suspect it would make sending outgoing messages much slower in scenarios where you want to send lots of messages as quickly as possible:
```csharp
using var scope = new RebusTransactionScope();

var headers = new Dictionary<string, string> { [Headers.Express] = "true" };

foreach (var message in lotsOfMessages)
{
	await bus.Send(message, headers);
}

await scope.CompleteAsync();
``` 

If you want it to work this way for you, I suggest you create an extension method on `IBus` that hides the details, e.g. something like 
```csharp
public static class ExpressBusExtensions
{
	public static async Task SendExpress(this IBus bus, object message)
	{
		using var scope = new RebusTransactionScope();

		var headers = new Dictionary<string, string> { [Headers.Express] = "true" };

		await bus.Send(message, headers);

		await scope.CompleteAsync();
	}	
}


``` 
I hope that makes sense 🙂

PR #1066, opened 2022-11-30
Change the default back off strategy to include some jitter of 0-9ms each time

kendallb:
We have run into a strange problem with the MySQL transport which I believe is a bug in MySQL itself (or the Percona XtraDB cluster that we use) and I have a solution for that here:

https://github.com/rebus-org/Rebus.MySql/pull/25

So far, it looks like I am right as I have yet to see the problem again. Basically I suspect MySQL is sometimes incorrectly handling a FOR UPDATE lock on a record and acting as though it was used with LOCK IN SHARE MODE. The upshot of this is that if we end up with a lot of threads all trying to snag a message at the exact same time, the first thread in will get the message and then update it in the transaction, and once it's transaction is completed the other message handlers are supposed to get the next message. This seems to work 99.99% of the time. But sometimes a bunch of threads all snag the same message at once. If we had use LOCK IN SHARE MODE then when the other threads wake up they would read the same message with updated values, which is not what we want. We want them to get the next message. FOR UPDATE is supposed to protect against that, but sometimes it breaks. Just yesterday we actually had a message get processed 16 times. We have three Rebus message queues running on three separate nodes, all with 10 threads each. So we have 30 threads processing messages. So it was quite odd to see that somehow 16 threads all happened to align at the same time and get the same message. I suspect my fix if I am right will work, as I check if the message that is read was actually already leased by another handler, and if so we just bail and try to get another one.

Anyway, what does this have to do with this PR? Quite simple it got me wondering how exactly we ended up with 16 threads all perfectly aligned and trying to poll messages at the same time? You would think that when the message queue has stuff to do, over time they would all get out of alignment as each message is going to take different amounts of time to run. But we have 30 threads total and our message throughput is actually pretty low. 30 threads is probably way over kill at the moment, but we have set it up for future growth as we convert more stuff over to messages. So in our case, a lot of the time the threads are actually going to end up sleeping and running through the back off strategy. The default is to sleep 100ms at a time until we have been asleep for 10 seconds, and then it always sleeps for 250ms. Given that you would think all the threads should over time diverge in timing, it is a bit weird that 16 of them all suddenly aligned?

Well I think it actually makes sense, because if a thread tries to get another message while one is being in the process of being read, it will block until it can get the next one. If another thread also happens to come along and do that, it will block also etc. What we have seen is that these issues seem to happen during times of heavy load, so when there are more messages in the queue to be processed a lot of the threads are likely not backing off and keep looping. In the process of looping, many of them will be blocking at the same deadlock point for tiny fractions of time and hence slowly get aligned. 

So I think the better help with getting all the threads to diverge in timing, we should introduce some jitter by defaults into the default back of strategy. It would not need to be much, so in this PR I set it up to be 0-9ms. Since the threads should theoretically be coming into the back off code and slightly different times, I figured it would make sense to use the current delay time millisecond count and just do a mod 10 on it, and add that to the timing to add 0-9ms of relatively random jitter. So that is what this PR does. The result of this is that when threads are backing off, they should theoretically all back off with slightly different delays which should in practice also mean the average latency for a thread to wake up and process a message is better than the worst case 250ms (or regular 100ms) due to the thread jitter. If they all get aligned which is what seems to be the case I am seeing then when they go to sleep they all going to be waking up at exactly the same time which means the average latency is going to be pretty much dead on 100ms to 250ms rather than diverging slightly. Not only is that bad for average latency, it is also bad because it will put much higher burst load on the SQL connection (but that is probably to a big deal).

An option to consider to make it more 'random' would be to not use the modulus of the millisecond delay, but perhaps use a shared random number generator to generate a random number of milliseconds to delay. That would probably make it a lot more random as it is possible I guess for the threads if they are aligned to all have the exact same total milliseconds of delay and remain aligned even with the jitter?

---

### DIFF
--- a/Rebus/Workers/ThreadPoolBased/DefaultBackoffStrategy.cs
+++ b/Rebus/Workers/ThreadPoolBased/DefaultBackoffStrategy.cs
@@ -106,12 +106,16 @@ class DefaultBackoffStrategy : IBackoffStrategy
             Interlocked.Exchange(ref _waitTimeTicks, waitedSinceTicks);
         }
 
-        var waitDurationTicks = DateTime.UtcNow.Ticks - waitedSinceTicks;
-        var totalSecondsIdle = (int) TimeSpan.FromTicks(waitDurationTicks).TotalSeconds;
+        var waitTimespan = TimeSpan.FromTicks(DateTime.UtcNow.Ticks - waitedSinceTicks);
+        var totalSecondsIdle = (int)waitTimespan.TotalSeconds;
         var waitTimeIndex = Math.Max(0, Math.Min(totalSecondsIdle, _backoffTimes.Length - 1));
 
-        var backoffTime = _backoffTimes[waitTimeIndex];
+        // Now find the jitter delays by taking the millisecond value since we started waiting to get
+        // a jitter value from 0-9 milliseconds. This will give each thread a slightly different delay each time
+        // so all the threads over time end up spread out
+        var jitterMilliseconds = waitTimespan.Milliseconds % 10;
 
-        return backoffTime;
+        // Now return the total delay
+        return _backoffTimes[waitTimeIndex].Add(TimeSpan.FromMilliseconds(jitterMilliseconds));
     }
 }


kendallb:
Changed to use random number generator and pushed that.

kendallb:
Yeah that’s makes sense. I actually have a plug in back of strategy now but have not yet deployed it. Was waiting to see if the latest changes I made in the MySQL transport fixed the duplicate message processing and it looks like it has. So I will use this approach instead. Thanks!

Issue #1122, opened 2023-11-28
How to access raw exception in Rebus 8


diegodfsd:
Hi @mookid8000,

Are there any ways to access the raw exception from IFailed<T> or any other means in case of failure, considering we are upgrading to Rebus 8? Our current flow was created over Rebus recommendations where the raw exception is important data for us.

ladenedge:
How fortuitous: I actually have a similar question.

Previously we've used exception metadata to return status codes or determine severity.  For instance, we might have a `StatusCodeException` that is returned with `IFailed<MyMessage>` whose property `StatusCode` can help us generate the appropriate `AbortSessionRequest` message type to publish on the bus.  (Or something like that.)

I'd be happy to hear about better ways to do this, but in the meantime it seems like `ExceptionInfo` is lacking in a way to encode metadata such as exception properties or `Exception.Data` information.

I understand we want `ExceptionInfo` to be more serializable and generic to accommodate distributed error tracking, but it seems like a richer object is necessary, and better conversions to/from exceptions might be nice, too.  My current approach is to marshal `ExceptionInfo` into a [Problem Details](https://datatracker.ietf.org/doc/html/rfc7807) structure, with some custom "extensions" for my specific metadata needs.  I was in fact planning to suggest that ASP.NET's `[ProblemDetails](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.problemdetails)` might be a good alternative to `ExceptionInfo`.  (Or, if we don't want to include that namespace as a Rebus dependency, perhaps just a nice model for `ExceptionInfo` to mirror.  Specifically, the dictionary for extensions/metadata, and the fancy JSON serialization.)

With Rebus as it exists today, it looks like I have two practical options:

1. Do my own (de)serialziation to the Details property of `ExceptionInfo`.
2. Subclass `ExceptionInfo`, then implement my own `IErrorTracker` to populate `FailedMessageWrapper`. (Would this work?)

Your thoughts and suggestions are most welcome!

labeled:
We are also running into a similar issue where we used exception metadata to return information to our users. Say someone tries to modify data that's already been modified, we took the new data and sent it with the exception such that the user could be updated.

Iadenedge I've tried to (de)serialize the Details property but since it's an `Exception.ToString()` it only returns the message and the stack strace, all of the metadata from the custom exception is lost.

ladenedge:
[...]

PR #1124, opened 2023-12-05
Introduces an IExceptionInfoFactory to customize the ExceptionInfo class [Refs #1122]

ladenedge:
This PR introduces `IExceptionInfoFactory` to customize how `ExceptionInfo` classes are created from exceptions.  The main points are:

1. Classes converting exceptions into `ExceptionInfo`s now have a dependency on `IExceptionInfoFactory`.
2. The default exception info factory is just `ExceptionInfo.FromException()`, so out-of-the-box behavior should not change.
3. We also provide an `InMemExceptionInfo` (and corresponding factory) such that interested users can revert to an `ExceptionInfo` that contains the raw exception.  The new factory would have to be registered at bus configuration time to override the default.
4. Custom info users may then make use of `ExceptionInfo.ConvertTo<TInfo>()` to get their custom info.  (This is probably the kludgiest part of this approach. ☹️)

For the moment, this is just an idea to review.  If you like it, I will of course add full unit test coverage, and perhaps some configuration helpers.  If you don't like the idea, either whole or in part, please let me know how things (even tiny things) might be changed for the better.  I'll keep an eye on #1122 for discussion.

Thanks for your time!

---

### DIFF
--- a/Rebus.Tests.Contracts/Errors/ExceptionInfoTests.cs
+++ b/Rebus.Tests.Contracts/Errors/ExceptionInfoTests.cs
@@ -1,0 +1,71 @@
+﻿using System;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using Rebus.Retry;
+
+namespace Rebus.Tests.Contracts.Errors;
+
+public class ExceptionInfoTests<TExceptionInfoFactory> : FixtureBase where TExceptionInfoFactory : IExceptionInfoFactory, new()
+{
+    protected TExceptionInfoFactory _factory;
+
+    protected override void SetUp()
+    {
+        base.SetUp();
+        
+        _factory = new TExceptionInfoFactory();
+    }
+
+    [Test]
+    public async Task ThrowsOnNullException()
+    {
+        Assert.That(() => _factory.CreateInfo(null), Throws.TypeOf<ArgumentNullException>());
+    }
+
+    [Test]
+    public async Task TypeEqualsQualifiedExceptionType()
+    {
+        var info = _factory.CreateInfo(new Exception("a"));
+        Assert.That(info.Type, Is.EqualTo("System.Exception, mscorlib"));
+    }
+
+    [Test]
+    public async Task MessageIncludesExceptionMessage()
+    {
+        var info = _factory.CreateInfo(new Exception("a"));
+        Assert.That(info.Message, Contains.Substring("a"));
+    }
+
+    [Test]
+    public async Task DetailsIsNotEmpty()
+    {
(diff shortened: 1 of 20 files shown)


mookid8000:
This is IMO a rather elegant solution to your problem, and you have coded it like I would have 🤩 If you add a test or two, it would be even better, of course 😁 I'll merge it ASAP when you tell me it's ready

ladenedge:
Thanks for the encouragement.  I'll try to polish this up early next week.

ladenedge:
These commits do a couple things:

870b1dd : Full unit test coverage (I think) for the new ExceptionInfo-related classes, including the original one.  I tried not to be too smart -- for instance, they only check that `Details` is non-empty, not that it has to be from `ToString()`.  I also added a configuration helper off of `Errors(...)` that looks like this IRL:

```cs
Configure.With(activator)
    .Errors(e => e.UseInMemExceptionInfos())
```

ad10b67 : From what I can tell, the unit tests require a good number of manually-created test classes (fakes, I guess) to test some of these systems with many dependencies.  On the chance you haven't seen these glue libraries **AutoFixture.NUnit3** and **AutoFixture.AutoMoq** -- or, more likely, just haven't had the time to integrate them -- this commit demonstrates how a class with lots of DI, `DefaultRetryStep`, might be unit tested pretty easily with these tools.  You can see [some of these tests](https://github.com/rebus-org/Rebus/pull/1124/commits/ad10b67b8a0af2264d96d04f6bc4e4ee0bc1b8f3#diff-c850e3de8c2a50bf7474990f4e9cbe3d770e87a5d6adabbc88e539b1f9337eda) are done with only a few lines and no manually-written fakes because we rely on AutoFixture/Moq to simply give us a sensible anonymous `DefaultRetryStep` that we can play with.

Even tests that do require enumerating the dependencies, such as `CreatesInfoForSecondLevelRetryException`, are still pretty straightforward.  Unit tests like these can also encourage good IoC because it's easier to test!

If you like this strategy, perhaps it can be integrated more as time goes on.  Without it, I feel like `DefaultRetryStep` is much less fun to test. 🙁 However, I realize it does add some new dependencies which you seem to have worked hard to avoid, so if you don't like it I'm happy to revert this commit, no harm done.  Lemme know what you think!

mookid8000:
I don't have a problem with trying out some AutoFixture stuff in the test project – either it'll grow roots and be used in other places as well, or it'll be torn out some time in the future 😉 

Thanks for taking the time to contribute with this – it will be out as Rebus 8.1.0 on NuGet.org in a few minutes if the tests are 🟢

Issue #1127, opened 2023-12-20
New retrystrategy seems incompatible with Azure Service Bus Native deadlettering


hjalle:
Hi,

The old "SimpleRetryStrategy" never created a new RebusTransactionScope before dispatching the message to the errorHandler:
``` cs
async Task MoveMessageToErrorQueue(IncomingStepContext context, ITransactionContext transactionContext, Exception exception)
    {
        var transportMessage = context.Load<OriginalTransportMessage>().TransportMessage.Clone();

        await _errorHandler.HandlePoisonMessage(transportMessage, transactionContext, exception);
    }
``` 

The new DefaultRetryStrategy does:
```cs
    async Task PassToErrorHandler(StepContext context, ExceptionInfo exception)
    {
        var originalTransportMessage = context.Load<OriginalTransportMessage>() ?? throw new RebusApplicationException("Could not find the original transport message in the current incoming step context");
        var transportMessage = originalTransportMessage.TransportMessage.Clone();

        using var scope = new RebusTransactionScope();
        await _errorHandler.HandlePoisonMessage(transportMessage, scope.TransactionContext, exception);
        await scope.CompleteAsync();
    }
```

In Rebus.AzureServiceBus the transactionContext.Items are read, but since its a new transactioncontext with the new retrystep, nothing is found:
```cs
public async Task HandlePoisonMessage(TransportMessage transportMessage, ITransactionContext transactionContext, ExceptionInfo exception)
        {
            if (transactionContext.Items.TryGetValue("asb-message", out var messageObject)
                && messageObject is ServiceBusReceivedMessage message
                && transactionContext.Items.TryGetValue("asb-message-receiver", out var messageReceiverObject)
                && messageReceiverObject is ServiceBusReceiver messageReceiver)
            {
....
```
which in turn just "elses out" and uses the default error handler.

I'm not really sure what the preferred way to solve this would be, thus making an issue instead of a PR.

hjalle:
Any progress/ideas regarding this one?

AndreaCuneo:
@mookid8000 I hit the same issue when using ASB NativeDeadlettering and new error handling.

As far as I can see, the issue stem from the decision of the RetryStep to create a 'nested' TransactionContext to handle the error.
This makes sense if the ErrorHandling wants to Send a new Message to a separate Queue and thus Complete() that sub-transaction.

[...]

PR #1155, opened 2024-03-23
fix(Retry): do not create nested scope for ErrorHandler in DefaultRetryStep

AndreaCuneo:
ErrorHandler impl should create their own nested scope when needed.

Update the DeadletterQueueErrorHandler accordling.

fix: #1127 

---


AndreaCuneo:
@hjalle @mookid8000 I did not reproduced the ASB integration error after this change and not plan to. I checked the Rebus.ASB code and my 👀 are convinced that it will work. Another pair of 👀 would be great to double check nonetheless.

mookid8000:
Your PR is  out as Rebus 8.3.0 on NuGet org now! 🙂

PR #1173, opened 2024-08-30
Add FlattenHierarchy binding flag in ActivateHandlersStep::GetDispatchMethod

X509Whatever:
Hi!

The class **ActivateHandlersStep** has the method CreateHandlerInvoker marked as virtual (useful++).

Trying to **override** it, I noticed in runtime that the method to make _GetHandlerInvokers_ as generic method was failing because wasn't found.

image

It could be great if it includes the binding flag flatten hierarchy so that when we extend this class we don't have to copy this method again (being able to use the original one).

Regards

---


mookid8000:
A small, but pretty excellent improvement 😎 thanks! It's out in Rebus 8.4.4 on NuGet.org now

PR #1175, opened 2024-09-03
Add coverage for inherited ActivationHandlersStep plus a missing visibility change

X509Whatever:
In the PR #1173 I forgot something :(
If the method to be reflected in the (base) extended class is private, adding the BindingFlag FlattenHierarchy does not work.

Here I try to fix it with a regression test (trying to follow the codestyle I have been seeing).
Regards.

---


mookid8000:
Excellent work! 👍

PR #1178, opened 2024-09-18
Fix leaks & performance tweaks

Henr1k80:
- Dispose some disposables
- Avoid allocations of empty arrays
- Use ValueTuples instead of Tuples & anonymous classes  
Less allocations, performance much faster in 8+ and around the same performance in Framework
- More static
- Other minor tweaks™®©

---


mookid8000:
Cool, thanks! 👍 Rebus 8.6.0 has the changes and is on Nuget.org now 🙂

Issue #1212, opened 2025-06-19
DefaultRetryStrategy, TransactionScope and EnlistInAmbientTransaction problem?


hjalle:
Hi,

I believe there's an issue in Rebus v8 when you are combining the defaultretrystrategy, TransactionScope and SQL provider enlisting in the ambient transaction.

I believe that I've reproduced the issue inside the SqlServer-repo, but only if i pull in the transactionscope dependency.

```cs
public class SqlTransactionsSopeError : FixtureBase
{
    static readonly string ConnectionString = SqlTestHelper.ConnectionString;

    BuiltinHandlerActivator _activator;
    IBusStarter _starter;

    protected override void SetUp()
    {
        DropTables();

        _activator = new BuiltinHandlerActivator();

        Using(_activator);

        _starter = Configure.With(_activator)
            .Transport(x => x.UseSqlServer(new SqlServerTransportOptions(ConnectionString, true), "test.input"))
            .Sagas(x => x.StoreInSqlServer(ConnectionString, "Sagas", "SagaIndex"))
            .Options(x =>
            {
                x.HandleMessagesInsideTransactionScope(new TransactionOptions()
                {
                    IsolationLevel = IsolationLevel.ReadCommitted
                });
                x.SetNumberOfWorkers(1);
                x.SetMaxParallelism(1);
            })
            .Create();
    }

    protected override void TearDown()
    {
        DropTables();
    }

    static void DropTables()
    {
        SqlTestHelper.DropTable("RebusMessages");
        SqlTestHelper.DropTable("SagaIndex");
        SqlTestHelper.DropTable("Sagas");
    }

    [Test]
    public async Task WaitForExceptions()
    {
        var receivedMessageCount = 0;

        var count = 0;
        _activator.Handle<string>(async _ =>
        {
            count++;
            throw new Exception();
        });

        var bus = _starter.Start();
        await bus.SendLocal("hej med dig min ven!");
         

        await Task.Delay(TimeSpan.FromSeconds(10));
        Assert.That(count, Is.EqualTo(5), "Only 5 deliveries should have been made");
    }
}
```

The test above fails as more than 5 deliviers have been made. The inner transactionscope is aborted which makes it impossible to commit the "read" of the message itself, thus leaving it forever running. The old RetryStep did not handle this in the same way - it started with the errorhandler logic, thus removing the need to have a failed message before removing the message (thus avoiding the abandoned context).

I wonder if thats the way forward in this case as well, essentially moving 
```cs
[...]

PR #1213, opened 2025-06-19
DefaultRetryStep failure handling moved to before processing message …

hjalle:
In reference to #1212, this is what I believe might be a solution, to avoid destroying the transactionscope that may be enlisted to.

---


hjalle:
@mookid8000 any possibilities for something like this?

mookid8000:
hi @hjalle , sorry for the long silence here 😅 

I have been pondering over this particular thing in the back of my mind since you first wrote, and I still really haven't reached a conclusion. Maybe we can talk about it here and somehow approach something together?

The thing is this: All versions of Rebus < 8 had the error tracker check BEFORE dispatching the message, which was originally so because of MSMQ message queue transactions. From Rebus 8, all transports have worked the same way: 

```
1. Try to handle message – buffer all outgoing messages in memory

2. If successful: Send outgoing messages and ACK incoming message
   If it fails: 
       If it hasn't failed too many times: NACK incoming message
       It it has failed too many times: Move message to error queue and ACK incoming message
```

BUT as you have discovered, transaction scopes and the SQL transport cause this model to fail, because it effectively emulates the old behavior of MSMQ where all the queue interactions are enlisted in the same transaction.

Honestly, I am more inclined to remove the transaction scope support, because it messes with the _try-to-handle-message-ACK-or-NACK_-model.

What are your thoughts?

hjalle:
Yes, I do see the point, but its kind of frustrating right now. We, and others I presume, have not yet fully migrated away from "DTC"-thinking.

Our current workaround is essentially that we decorate the SQL-transport and have a in-memory delivery count, which will then "fail early" in the DefaultRetryStep, as that logic resides at the top.. I'm not sure if that has any drawbacks but at least it solves the issue with endless retry-loops. I don't really see it as a viable solution right now since it relies on the internals of the DefaultRetryStep that one as well, although it of course makes a lot more sense to have that logic at the beginning of the retry rather than at the end of the "last" delivery. Its also not a generic solution as v8 with enlistInAmbient, sql and transactionscopes will result in endless retries for all existing users. If the SQL-transport would add delivery counting, it would probably solve this issue, I guess..?

GMacAussie:
First off, we are still on Rebus 6 because we have our own outbox/transaction scope handling and didn't want entanglement, so therefore this comment will be most likely completely offbase...

We have had in the past the need to separate out (isolate) transaction scopes (for various reasons). We do that using:

```
        /// <summary>
        /// We usually use Read Committed isolation and default timeout
        /// </summary>
        static readonly TransactionOptions ReadCommitted = new TransactionOptions {
            Timeout = TransactionManager.DefaultTimeout,
            IsolationLevel = IsolationLevel.ReadCommitted
        };

        /// <summary>
        /// Create a suppress read committed transaction scope. By default the isolation level is Serializable
        /// which is not the database uses, which is Read Committed.
        /// 
        /// A suppress transaction scope is used to separated transaction scope from an existing transaction
        /// scope. It will not be an ambient transaction scope.
        /// 
        /// Make sure TransactionScopeAsyncFlowOption is enabled to allow the transaction scope to cross async/await.
        /// </summary>
        /// <returns>The suppress read comminited transaction scope</returns>
        public static TransactionScope CreateSuppressReadCommittedScope() {
            return new TransactionScope(TransactionScopeOption.Suppress, ReadCommitted, TransactionScopeAsyncFlowOption.Enabled);
        }
```

Just throwing this out there...

